PR #2, opened 2024-05-10
Migrate to the FHIRPath service

karlnaden:
*** WORK IN PROGRESS ***

tested with https://github.com/inferno-framework/fhirpath-service/tree/fi-2700-fhirpath-service and seems to work, however

1. previously, when fhirpath evaluation failed, Inferno continued assuming it returned nothing
2. I haven't found a scenario where having the service down changes the output. 
I'd like to find such a scenario and validate it to make sure that fhirpath eval is actually effective.


karlnaden:
Done in https://github.com/inferno-framework/davinci-pas-test-kit/pull/4

PR #3, opened 2024-05-24
More gracefully handle missing claim and missing fullUrl

tstrass:
# Summary

Issue: In the client suite, we were return a 500 "low level error occurred" in the event that a claim submission was missing a claim or if the claim was missing a fullUrl (which are both required in the spec). The claim validation test also resulted in an inferno error if there was no fullUrl.

Resolution: Add a check to avoid a 500, return a 400 with an appropriate OperationOutcome instead. Also add safe navigation in the validation module to avoid an inferno error in the validation test.

# Testing Guidance

Run a $submit test. Send it a bundle without a claim. Sent it a bundle with a claim that is missing a fullUrl.

### DIFF
--- a/lib/davinci_pas_test_kit/custom_groups/v2.0.1/client_tests/pas_client_approval_submit_response_attest.rb
+++ b/lib/davinci_pas_test_kit/custom_groups/v2.0.1/client_tests/pas_client_approval_submit_response_attest.rb
@@ -20,6 +20,7 @@ module DaVinciPASTestKit
             )
 
       run do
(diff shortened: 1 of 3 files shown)


karlnaden:
in the case that a 400 error is returned, the attestation test should skip rather than ask for the attestation since we didn't send the actual response back.

karlnaden:
skip update looks good. Will defer unit tests to later (see [FI-2761](https://oncprojectracking.healthit.gov/support/browse/FI-2761))

PR #4, opened 2024-05-31
FI-2701: Migrate to HL7 validator wrapper

dehall:
# Summary
This PR migrates the test kit from the Inferno validator-wrapper to the HL7 validator-wrapper. This is similar to most of the other test kits, with one significant difference that his test kit used the inferno validator's FHIRPath endpoint, so this now includes the fhirpath-service in the docker-compose files and nginx.background.conf.

Other changes:
- add the hl7_validator service to docker-compose and nginx. The old validator components are left commented out in case someone really wants the old validator UI
- Bump the core dependency
- Migrate `validator` blocks to `fhir_resource_validator` and specify the IG
- Update the generator with ^ and re-generate the suite
- Update the fhirpath evaluation method to call the appropriate endpoint (based on #2 )
- Add a README to the IGs folder explaining when it's necessary for the files to be there
- Fixes an issue where the custom validation logic tries to validate against native profile urls with the IG's version, in particular `http://hl7.org/fhir/StructureDefinition/PractitionerRole|2.0.1` . The validator doesn't have this profile loaded which results in a 500 error and a purple test result
- Updates the preset and sample bundles to remove relative links, eg `<a href="#something">` . A new check in the validator has strict requirements for these and produces "Hyperlink does not resolve" errors. The replacement was `<a href=\\\\\\"#.*?">See above \((.*?)\)</a>` --> `\1`, eg, `<a href=\\\"#Patient_SubscriberExample\\\">See above (Patient/SubscriberExample)</a>` became just `Patient/SubscriberExample`

# Testing Guidance
From a user's perspective there should be no visible difference between this and using the old validator. Run through the tests and make sure everything works


dehall:
Failing on the old one and passing on the new one is unexpected. I can look more into that

tstrass:
Here's my test run: https://inferno.healthit.gov/suites/davinci_pas_server_suite_v201/b8L0hB6WlyK#davinci_pas_server_suite_v201-Group01-pas_server_v201_approval_use_case

dehall:
I wonder if the prod validator instance may have been in an inconsistent state somehow when you tried it before, because test 1.1.1.01 with the preset input is passing now https://inferno.healthit.gov/suites/davinci_pas_server_suite_v201/cvEQvFrwC2p#1.1

dehall:
Ignore the previous (now-deleted) message if you saw that. I wonder if the prod validator instance is in an inconsistent state because running that same test locally on main passes for me:
image

dehall:
Ok I see the difference now. That error we're seeing on prod is for a CRD extension http://hl7.org/fhir/us/davinci-crd/STU2/StructureDefinition-ext-coverage-information.html and so the validator will only check that when the CRD IG is loaded, either explicitly or as a dependency of another IG. When you run this test kit by itself on main, you don't get those errors. But on prod the old validator instance is running for a few test kits and one of them may be bringing in a different version of CRD. If you load all the same IGs into the validator locally as are running on prod, the same error happens. (davinci-pdex-2.0.0.tgz, davinci_pas_2.0.1.tgz, uds-plus-external-package.tgz, davinci_dtr_2.0.1.tgz, udsplus-101.tgz)

So, I think this should be good, in terms of matching the old behavior when running in isolation. Let me know if you disagree @tstrass or @karlnaden , otherwise I'll merge this in

tstrass (approved):
Agreed

dehall:
Just for posterity, turns out that constraint that was failing is bad anyway. https://jira.hl7.org/browse/FHIR-46087 (thanks fhir-at-mitre slack channel)

PR #15, opened 2024-08-07
create unique identifier for pended test wait continuation

### LINKED ISSUE #13: resume for pended workflow is not session specific
The link to continue the pended workflow test for the PAS server tests uses a wait that looks for a generic request that is not specific to the session (see https://github.com/inferno-framework/davinci-pas-test-kit/blob/main/lib/davinci_pas_test_kit/custom_groups/v2.0.1/notification/pas_subscription_notification_test.rb#L20). This means that two concurrent testers could disrupt each other's sessions.

This should be updated to include information specific to the session.

karlnaden:
# Summary

Resolves [issue 13](https://github.com/inferno-framework/davinci-pas-test-kit/issues/13) where the wait dialog for the temporary subscription placeholder was not session specific, meaning that concurrent sessions could interrupt each other. The wait dialog now generates a unique key that is used which will be unique to the session.

# Testing Guidance

Create a PAS Server suite session and run group 1.3 Successful Pended Workflow using the Prior Authorization Support Reference Implementation preset. Verify that when the wait dialog appears, 
1. hovering over the "click here" link shows a url with a random "token" parameter at the end 
2. clicking on the "click here" link successfully continues the test.

### DIFF
--- a/lib/davinci_pas_test_kit/custom_groups/v2.0.1/notification/pas_subscription_notification_test.rb
+++ b/lib/davinci_pas_test_kit/custom_groups/v2.0.1/notification/pas_subscription_notification_test.rb
@@ -15,17 +15,17 @@ module DaVinciPASTestKit
       )
 
       run do
-        identifier = SecureRandom.hex(32)
+        token = SecureRandom.hex(32)
         # rubocop:disable Layout/LineLength
         wait(
-          identifier:,
+          identifier: token,
           message: %(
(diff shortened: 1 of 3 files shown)


vanessuniq (approved):
Work as expected

arscan (approved):
Tried it out and works as advertised.

When I mucked with the token it didn't find the route as expected.

Subsequent test runs changed the token as expected.

PR #17, opened 2024-09-18
Auth info

vanessuniq:
# Summary
Update to use auth info  and fhirpath evaluator feature from core.

# Testing Guidance
bundle install and run server tests locally.  Ensure the auth credentials display correctly and validation tests using fhirpath works. See more details about core [FHIRPath evaluator](https://inferno-framework.github.io/inferno-core/docs/Inferno/DSL/FhirpathEvaluation.html#evaluate_fhirpath-instance_method), [AuthInfo](https://inferno-framework.github.io/inferno-core/docs/Inferno/DSL/AuthInfo.html)

Screenshot 2024-09-25 at 10 41 08 AM


lib/davinci_pas_test_kit/generator/templates/suite.rb.erb
```diff
@@ -47,12 +47,15 @@ module DaVinciPASTestKit
 
       input :smart_credentials,
             title: 'OAuth Credentials',
-            type: :oauth_credentials,
+            type: :auth_info,
+            options: {
+              mode: 'access'
+            },
             optional: true
```
tstrass, line 54: Any way to make this just say "Auth Type" instead of "smart_credentials Auth Type"?

tstrass (changes requested):
~This is not related to this PR, but I noticed that suite generation is slightly out of date (non-functional diffs in 4 files), so might make sense to just bring it up to date while we have an open PR.~

Edit: Nevermind, this'll be a separate ticket

lib/davinci_pas_test_kit/generator/templates/suite.rb.erb
vanessuniq: it will look better. That's a question for @AlyssaWang . I believed the way it was implemented was bay concatenation the the input name and 'Auth Type'. This change will happen core.

tstrass: Gotcha, wasn't sure if it needed a core change, so thanks for confirming.

tstrass (approved):

karlnaden:
Trying to read through the linked details on the new `auth_info` input type and having trouble due to formatting issues on the class page - is that something that is being addressed in core, or should I create a ticket?

karlnaden:
Is the `Public` Auth Type the right default for the PAS server tests? The PAS connection seems more like a backend services type connection, but I don't know enough about SMART yet to be certain? Do you have a sense @vanessuniq and if you agree, could the default selection be updated?

vanessuniq:
> Is the `Public` Auth Type the right default for the PAS server tests? The PAS connection seems more like a backend services type connection, but I don't know enough about SMART yet to be certain? Do you have a sense @vanessuniq and if you agree, could the default selection be updated?

Yes, backend services seems more appropriate. I updated to make backend services the default.

karlnaden:
Selecting the reference implementation preset enables some strange additional details in the auth info input. Can this be cleaned up in this test kit, or is a core issue?

image

karlnaden:
> > Selecting the reference implementation preset enables some strange additional details in the auth info input. Can this be cleaned up in this test kit, or is a core issue?

> This is the correct behavior. Depending on the auth type and mode, different input fields need to be displayed. The extra fields are the fields the FHIR client may need to automatically refresh the token or request another access token in case the current one has expired. 

sounds good - thanks for explaining.

karlnaden:
the preset for the RI server still uses the old `outh_credentials` type for the `smart_credentials` input. Looks like it doesn't cause a functional problem, but should still be updated to match.

karlnaden (approved):
update the preset type and its good

PR #21, opened 2024-11-25
FI-3301: SuiteEndpoints

tstrass:
# Summary

Replace record_response_route monkey patch with SuiteEndpoints

# Testing Guidance

Run through the whole Client Suite and ensure the same results as before. This PR is a pure refactor and should result in no functional changes.


karlnaden (approved):
post man demo works as expected, code looks good

PR #28, opened 2025-03-11
FI-3818: Add PAS Client verifies_requirements

elsaperelli:
# Summary
This PR adds `verifies_requirements` annotations to associated tests in the PAS Client test suite. Any requirements that are not currently tested were added to a Jira ticket for attestation test creation.

# Testing Guidance
- `bundle exec rspec`
- `bundle exec rubocop`
- Run the PAS Client test suite
- Go through the `verifies_requirements` and make sure the test matches the requirement and that it is at the most refined level
- Make sure none of the requirements listed in the Jira task for attestation testing do not have associated tests already implemented


karlnaden:
Notes - 
- 58, 67, 119, 153 - all client requirements associated with server tests, but can be moved
    - 58 all workflow submit tests
    - 67, 119, 153 pended workflow submit test only
- @75 - should be tested - see pas_bundle_validation.rb line 83
- @120 - covered by the pended submit test
- @202 and @203 (submit and inquire support) should be added to the wait tests (only the pended one for inquire)
- @155 not tested (I also decided it was a server requirement, but it will be attestation)
- @62 - would put on the wait test, not the validation
- @80 and @83 - I don’t think we explicitly test
- @144 - we don’t currently require id-only
- @166, 169, 170 - we don’t test update scenarios yet

karlnaden (changes requested):
requested some mapping updates

karlnaden (approved):

PR #29, opened 2025-03-18
FI-3828: Add PAS Server verifies_requirements

elsaperelli:
# Summary
This PR adds `verifies_requirements` annotations to associated tests in the PAS Server test suite. Any requirements that are not currently tested were added to a Jira ticket for attestation test creation.

# Testing Guidance
- `bundle exec rspec`
- `bundle exec rubocop`
- Make sure `bundle exec rake requirements:generate_coverage` has been run
- Run the PAS Server test suite
- Go through the `verifies_requirements` and make sure the test matches the requirements and that it is at the most refined level
- Make sure none of the requirements listed in the Jira task for attestation testing do not have associated tests already implemented


karlnaden:
- @5 can be associated with the exchange tests for submit and inquire interactions - they are exchanging a FHIR bundle
- @35 tested by 2.1.2.03 - if that’s generated, we probably want to associate it with the 2.1.2 group
- @36 tested by 2.2.2.03 - if that’s generated, we probably want to associate it with the 2.2.2 group
- @66,@136 tested by the pended workflow, maybe associate with 1.3.1.02?
- @110 - I think tested by the ClaimResponse must support tests because the error fields are all must support
- @33 - HRex must support requirements - add to the top-level must support group(2)

karlnaden (approved):

PR #32, opened 2025-03-26
Add verifies_requirements to the generator

elsaperelli:
# Summary
This PR adds `verifies_requirements` annotations to the generator files so that the annotations are included when `pas:generate` is run. I also made some formatting changes to the generator, but was unable to fix all of them, hence some of the diffs.

# Testing Guidance
- `bundle exec rspec`
- `bundle exec rubocop`
- `bundle exec rake pas:generate`
- `bundle exec rake requirements:generate_coverage`


tstrass (changes requested):

.github/workflows/ruby.yml
```diff
@@ -21,12 +21,36 @@ jobs:
...
+        - name: Set up Ruby
+          uses: ruby/setup-ruby@v1
+          with:
+            ruby-version: ${{ matrix.ruby-version }}
+            bundler-cache: true
+        - name: Verify test generation is up to date
+          run: |
+            bundle exec rake us_core:generate
+            if [[ -n "$(git status --porcelain lib/us_core_test_kit/generated)" ]]; then
+              echo "Generated tests are outdated. Run 'bundle exec rake us_core:generate' and commit the changes."
```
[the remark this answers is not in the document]
elsaperelli: Ooops!

lib/davinci_pas_test_kit/generator/templates/group.rb.erb
```diff
@@ -8,7 +8,7 @@ module DaVinciPASTestKit
         <%=description.gsub("\n", "\n" + " "*8) %>
       )
 
-      id :<%= group_id %>
+      id :<%= group_id %><%=verifies_requirements%>
```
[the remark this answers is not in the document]
elsaperelli: No this makes sense! I will change this over. I did have this at one point but was hesitant about those new lines created but makes sense that that would be nbd and this would be cleaner code.

lib/davinci_pas_test_kit/generator/group_generator.rb
```diff
@@ -217,12 +217,16 @@ def grouped_must_support_test_ids
...
             'Submit Response Must Support' => submit_response_must_support_test_ids
           },
           '$inquire Element Support' => {
             'Submission of claims to the $inquire operation for must support validation' => inquiry_operation,
-            '[USER INPUT VALIDATION] Inquiry Request Must Support' => inquiry_request_must_support_test_ids,
+            "[USER INPUT VALIDATION] Inquiry Request Must Support'" \
+            "\n#{' ' * 10}verifies_requirements " \
+            "'hl7.fhir.us.davinci-pas_2.0.1@36" => inquiry_request_must_support_test_ids,
             'Inquiry Response Must Support' => inquiry_response_must_support_test_ids
           }
```
tstrass, line 231: I'm not clear on what this is doing. Can we get rid of it now, or is this another spot where the `verifies_requirements` generation needs to be updated?

tstrass (changes requested):

lib/davinci_pas_test_kit/generator/group_generator.rb
elsaperelli: Unfortunately this is another spot where `verifies_requirements` generation needs to be updated. I added a `verifies_requirements` to `pas_server_must_support_use_case_group.rb` that gets removed if I get rid of this code when running `bundle exec rake pas:generate`. Happy to reconsider this approach, but I think we need it if we don't want mismatches when generate is run.

tstrass: Leaving this as is. It's not great shoehorning in the verifies_requirements into the title string, but it would require more refactoring than it's worth

tstrass (approved):

PR #33, opened 2025-04-02
FI-3813: Use core MustSupport features

dehall:
# Summary
This PR updates the Must Support tests to use the core MS logic -- `MustSupportMetadataExtractor` in the metadata generation phase and `assert_must_support_elements_present` in the test. All the metadata and tests have been regenerated.

This migration revealed an issue in the core logic related to type slicing on a Bundle so those changes are in https://github.com/inferno-framework/inferno-core/pull/659 which has now been released as `0.6.9`

# Testing Guidance
Unfortunately the way the UI is built in core means that pointing this test kit to a branch breaks the UI. So until the core PR is merged and released - see instructions on that core PR.

Run the test suite against the preset. All results should either match the current test on prod, or where there are differences, the new result should be correct on closer inspection of the data.

Regenerate the test suites and make sure there are no differences. I recommend running as follows to make sure nothing gets missed:
```
rm -r lib/davinci_pas_test_kit/generated/
bundle exec rake pas:generate
git diff
```


tstrass (commented):
I'm seeing one difference between the results on main and this branch. This test is passing on main:
image

I've inspected it, and I believe the new test result is correct.

Preliminary ✅ from me.

tstrass (approved):

PR #35, opened 2025-04-28
FI-4003: Initial attestation->demonstration tests

elsaperelli:
# Summary
This PR converts some of the existing attestation tests for PAS Client into demonstration or mechanical checks.

## New Tests:
- Requirements 11, 187, 188:
    - I added a `pas_client_review_submit_atttest.rb` file which includes a test for checking that the client permits, but does not require, the provider to reveiw the prior authorization submission data prior to transmission. This test was added to the `$submit` workflow tests: `pas_client_approval_group.rb`, `pas_client_denial_group`, and `pas_client_pended_group.rb`. 
     - These requirements talk about submission, so I didn't think they needed to be added to `$inquire` tests.

- Requirements 45, 47, 49:
    - I added `us_core_test_kit` as a dependency in the `.gemspec` file so that I could bring in test groups for US Core R4 profiles Condition, Observation, and Procedure. I used v3.1.1 tests, but not sure if that is correct. Also, there didn't seem to be a singular Observation test group, so I brought in all the Observation test groups.

## Questions about other tests to be added:
- 159, 164, 166-175:
    - I definitely wanted to discuss these more before diving in, are we thinking about adding new workflows for cancellation and modification (like we have for approval, denial, and pended)? If so, definitely would like to talk through what that should look like.
- 204-206: 
    - Do these need to be added in the subscriptions test kit?
    - I saw a read_endpoint in the subscription test kit, maybe that can be pulled in?
- 80, 83:
    - I spent a lot of time looking into this, but I am not totally sure what the best path is or what the spec even means by `Claim.supportingInfo.valueReference`. I doesn't _look_ to me like that exists on the [PAS Claim Profile](https://hl7.org/fhir/us/davinci-pas/StructureDefinition-profile-claim.html), unless it is referring to `Claim.supportingInfo:AdditionalInformation.value[x]`? I didn't think that was the same thing but maybe it is? If it is, I still have additional questions about how this could be tested since the must support tests in this test kit are generated with `metadata.yml`...
- 144:
    - Does this need to be added directly to the subscriptions test kit?

# Testing Guidance
- `bundle exec rspec`
- `bundle exec rubocop`
- `bundle exec rake requirements:generate_coverage` 
- `bundle exec inferno services start`
- `bundle exec inferno start`
- Go through the tests by following the instructions under "Postman-based Demonstration". Note: I have not updated this documentation yet to include the new tests, I want to make sure they will be added first.


karlnaden:
Wins are not as clear as hoped - closing

PR #36, opened 2025-05-01
FI-4053: Suite options for client auth

karlnaden:
# Summary

Use suite options to determine the authentication / session identification approach and clean up the inputs so only those used in the selected option are shown.

# Testing Guidance

Follow the instructions on the suite documentation to test with postman.


tstrass (changes requested):
Minor stuff. The only important thing is regenerating requirements coverage.

This was a little tricky to review since there's so much moved code, but it's looking good to me and I couldn't find any issues testing it out

tstrass (approved):

PR #52, opened 2025-09-19
Add presets for running the client against the server tests

karlnaden:
# Summary

Presets for running the client server suites against each other can be helpful for development purposes. They have been added with documentation.

# Testing Guidance

Follow the documented steps to execute the suites against each other.


karlnaden:
added to [PR 53](https://github.com/inferno-framework/davinci-pas-test-kit/pull/53)

PR #56, opened 2026-02-16
Id 54 v220 full resource subscription pended

abhinandan2012:
### Summary
Updates the PAS v2.2.0 pended workflow to use `full-resource` subscription notifications instead of `id-only`, and removes the `$inquire` operation from the pended workflow. In v2.2.0, the notification contains all details about the finalized claim, so the client no longer needs to perform a follow-up `$inquire` request.

#### Subscription Verification (Cross-Suite)
- `pas_subscription_verification.rb`: Added v2.2.0 validation logic for PAS Subscription profile (`meta.profile`), `rest-hook` channel type, `full-resource` payload content, and `org-identifier=*` criteria format.

#### PAS Notification Verification (New - Cross-Suite)
- `pas_notification_verification.rb`: New module that validates notification Bundles contain focus resources referencing the PAS Response Bundle profile.
- `pas_notification_conformance_test.rb`: New shared test class used by both client and server pended workflow groups.

#### Client Pended Group (v2.2.0)
- `pas_client_pended_group.rb`: Removed the `$inquire` group entirely. Added the PAS notification conformance test to the notification verification group. Added a "Verify final decision from notification" attestation group.
- `pas_client_pended_submit_test.rb`: Removed `inquire_json_response` input, `inquire_respond_with` configuration, and inquiry-related wait/response logic. The workflow now ends after the full-resource notification step.

#### Server Subscription Input Conformance (v2.2.0)
- `pas_server_subscription_input_conformance.rb`: Changed payload content assertion from `id-only` to `full-resource`. Passed `ig_version: 'v2.2.0'` to `verify_pas_subscription`.

#### Server Pended Use Case Group (v2.2.0)
- `pas_server_pended_use_case_group.rb`: Changed `subscriptions_r4_server_id_only_conformance` to `subscriptions_r4_server_full_resource_conformance`. Updated `subscription_type` option to `full-resource`. Added PAS notification conformance test. Removed all `$inquire` request validation, operation, and response validation tests. Removed the `claim_inquire_operation_test` require.

### Test Coverage
- `pas_client_subscription_pas_conformance_test_spec.rb`: Added v2.2.0 test context verifying passes with valid v2.2.0 subscription and fails with missing PAS profile, wrong channel type, and id-only payload.
- `pas_notification_conformance_test_spec.rb`: New spec validating PAS notification conformance – passes with focus resource, skips with no notifications, fails when missing focus.
- New fixtures: `PAS_Subscription_example_v220_good.json`, `PAS_Subscription_example_v220_bad.json`

### Backward Compatibility
- v2.0.1 behavior is fully preserved – the subscription verification module branches on `ig_version`
- All existing v2.0.1 tests remain unchanged and continue to pass

---

## Files Changed

| Status | File |
|--------|------|
| Modified | `lib/davinci_pas_test_kit/cross_suite/pas_subscription_verification.rb` |
| Modified | `lib/davinci_pas_test_kit/client/v2.2.0/pas_client_pended_group.rb` |
| Modified | `lib/davinci_pas_test_kit/client/v2.2.0/workflows/pas_client_pended_submit_test.rb` |
| Modified | `lib/davinci_pas_test_kit/server/v2.2.0/pas_server_subscription_input_conformance.rb` |
| Modified | `lib/davinci_pas_test_kit/server/generated/v2.2.0/pas_server_pended_use_case_group.rb` |
| Modified | `spec/davinci_pas_test_kit/client/pas_client_subscription_pas_conformance_test_spec.rb` |
| New | `lib/davinci_pas_test_kit/cross_suite/pas_notification_verification.rb` |
| New | `lib/davinci_pas_test_kit/pas_notification_conformance_test.rb` |
| New | `spec/davinci_pas_test_kit/pas_notification_conformance_test_spec.rb` |
| New | `spec/fixtures/PAS_Subscription_example_v220_good.json` |
| New | `spec/fixtures/PAS_Subscription_example_v220_bad.json` |

---

### Objective
Validate that the PAS Test Kit correctly supports v2.2.0 `full-resource` subscription notifications in the pended workflow, including removal of the `$inquire` step, while maintaining backward compatibility with v2.0.1.

### Prerequisites
- Inferno Da Vinci PAS Test Kit running locally
- Branch `ID-54-v220-full-resource-subscription-pended`

### Validation Workflow Overview
1. Running the Client Suite v2.2.0 as a mock PAS server
2. Running the Server Suite v2.2.0 as the test client
3. Executing the pended workflow with full-resource notifications
4. Confirming that `$inquire` tests are no longer present in v2.2.0
5. Verifying backward compatibility with v2.0.1 behavior

---

### Step 1: Set Up Client Suite (TAB 1)
1. Open the Da Vinci PAS Client Suite v2.2.0
2. For "Client Security Type", select "Other Authentication"
3. Click "Start Testing" to create a testing session
4. From the Preset dropdown (top-left), select "Run Against the PAS Server Suite" – this auto-populates the input fields

### Step 2: Set Up Server Suite (TAB 2)
1. Open the Da Vinci PAS Server Suite v2.2.0
2. Click "Start Testing" to create a testing session
3. From the Preset dropdown (top-left), select "Run Against the PAS Client Suite" – this auto-populates the input fields

---

### Step 3: Verify Subscription Setup Tests (Client Suite v2.2.0)
Navigate to **1 Subscription Setup** in the left sidebar of the Client Suite v2.2.0.

1. Run the subscription setup tests
2. Confirm the subscription conformance test validates:
   - PAS Subscription profile (`http://hl7.org/fhir/us/davinci-pas/StructureDefinition/profile-subscription`) is declared in `meta.profile`
   - Channel type is `rest-hook`
   - Payload content type is `full-resource` (not `id-only`)
   - Filter criteria uses the `org-identifier=<identifier>` format

### Step 4: Verify Pended Workflow Structure (Server Suite v2.2.0)
Navigate to **2 Demonstrate Workflow Support > 2.3 Successful Pended Workflow** in the Server Suite v2.2.0 left sidebar. Expand it and verify the structure.

#### Expected v2.2.0 structure under "2.3 Successful Pended Workflow":

| Group/Test | Expected in v2.2.0 |
|---|---|
| **2.3.1 Server can respond with 'Pended' to claims submitted for prior authorization** | Present |
| 2.3.1.01 [USER INPUT VALIDATION] Provided $submit Request Bundle is conformant | Present |
| 2.3.1.02 Submission of a claim to the $submit operation succeeds | Present |
| 2.3.1.03 Server $submit Response Bundle is conformant | Present |
| 2.3.1.04 Server response includes the 'Pended' decision code in the ClaimResponse instance | Present |
| **2.3.2 Server can notify client of updates via full-resource notification** | Present (title changed from v2.0.1) |
| 2.3.2.01 Server notifies the client that the pended claim was updated. | Present |
| 2.3.2.02 Subscription Event Notification Verification | Present |
| 2.3.2.03 Subscription Full-Resource Notification Verification | Present (was "Id-Only" in v2.0.1) |
| 2.3.2.04 Notification conforms to PAS-specific requirements | **NEW** – PAS notification conformance |
| 2.3.2.04 [USER INPUT VALIDATION] Provided $inquire Request Bundle is conformant | **REMOVED** |
| 2.3.2.05 Submission of a claim to the $inquire operation succeeds | **REMOVED** |
| 2.3.2.06 Server $inquire Response Bundle is conformant | **REMOVED** |

**Key visual checks:**
1. The second group (2.3.2) title should say **"full-resource notification"**, NOT "respond to claims submitted for inquiry"
2. Test 2.3.2.03 should say **"Full-Resource"**, NOT "Id-Only"
3. There should be **NO $inquire tests** (no "Provided $inquire Request Bundle is conformant", no "Submission of a claim to the $inquire operation succeeds", no "Server $inquire Response Bundle is conformant")
4. A new test for **"Notification conforms to PAS-specific requirements"** should be present

### Step 5: Verify Pended Workflow Structure (Client Suite v2.2.0)
Navigate to the **Pended Workflow** group in the Client Suite v2.2.0. Expand it in the left sidebar.

#### Expected v2.2.0 structure:

| Group/Test | Expected in v2.2.0 |
|---|---|
| **Perform the pended workflow** | Present |
| (submit test – waits for $submit + notification only) | Present (no $inquire step) |
| **Verify $submit interaction** | Present |
| **Verify notification interaction** | Present |
| (includes PAS notification conformance test) | **NEW** |
| **Verify final decision from notification** | **NEW** – attestation for full-resource notification |
| Verify $inquire interaction | **REMOVED** |

**Key visual checks:**
1. There should be **NO** "$inquire" or "inquiry" group
2. There should be **NO** `inquire_json_response` input field
3. The submit test wait message should only mention two steps: (1) $submit and (2) notification - **no third step** for $inquire
4. A new "Verify final decision from notification" attestation group should be present

### Step 6: Verify the Server Subscription Input Conformance
Navigate to **1 Subscription Setup** in the Server Suite v2.2.0.

1. Run the subscription input conformance test
2. Open the test description - it should mention `full-resource` payload content type, NOT `id-only`

#### Server Suite v2.0.1
Switch to the **Da Vinci PAS Server Suite v2.0.1** and navigate to **2.3 Successful Pended Workflow**.

Confirm the **original v2.0.1 structure is unchanged**:

| Group/Test | Expected in v2.0.1 |
|---|---|
| **2.3.1 Server can respond with 'Pended' to claims submitted for prior authorization** | Present |
| 2.3.1.01 – 2.3.1.04 (submit tests) | Present |
| **2.3.2 Server can notify client of updates and respond to claims submitted for inquiry** | Present |
| 2.3.2.01 Server notifies the client that the pended claim was updated. | Present |
| 2.3.2.02 Subscription Event Notification Verification | Present |
| 2.3.2.03 Subscription **Id-Only** Notification Verification | Present |
| 2.3.2.04 [USER INPUT VALIDATION] Provided **$inquire** Request Bundle is conformant | **Still present** |
| 2.3.2.05 Submission of a claim to the **$inquire** operation succeeds | **Still present** |
| 2.3.2.06 Server **$inquire** Response Bundle is conformant | **Still present** |

#### Client Suite v2.0.1
Switch to the **Da Vinci PAS Client Suite v2.0.1** and navigate to the **Pended Workflow** group.
- Confirm the `$inquire` group is **still present**
- Confirm the `inquire_json_response` input is **still present**
- Confirm the subscription validation checks for `id-only` payload


lib/davinci_pas_test_kit/cross_suite/pas_notification_verification.rb
```diff
@@ -0,0 +1,46 @@
...
+        ))
+      else
+        # Check that at least one focus entry references a PAS Response Bundle profile
+        has_pas_response = focus_entries.any? do |entry|
+          resource = entry['resource']
+          next false unless resource
+
+          resource_profiles = resource.dig('meta', 'profile') || []
+          resource_profiles.any? { |p| p.start_with?(PAS_RESPONSE_BUNDLE_PROFILE) }
+        end
```
karlnaden: We can't rely on meta.profile since it is not a required field. Instead, you'll need to find the status entry, get the reference to the focus resource from it, and use that to find the appropriate notification Bundle entry. You'll also need to run the response Bundle through the already-defined Bundle validation methods.

lib/davinci_pas_test_kit/cross_suite/pas_subscription_verification.rb
```diff
@@ -1,32 +1,113 @@
...
+
+    def verify_v220_profile(subscription)
+      profiles = subscription.dig('meta', 'profile') || []
+      return if profiles.include?(PAS_SUBSCRIPTION_PROFILE)
+
+      add_message('error', %(
+          The Subscription must declare conformance to the PAS Subscription profile
+          `#{PAS_SUBSCRIPTION_PROFILE}`
+          in the `Subscription.meta.profile` element.
+        ))
```
karlnaden: Can't rely on the declaration in meta.profile to be present and couldn't rely on the assertion to be accurate, so need to run the conformance check against the profile using the validator.

karlnaden (changes requested):

lib/davinci_pas_test_kit/cross_suite/pas_notification_verification.rb
abhinandan2012: Updated - notification verification no longer relies on meta.profile. It now extracts the notification-event.focus reference from the SubscriptionStatus Parameters, resolves it to the corresponding Bundle entry, and validates the focus resource through the existing perform_response_validation method.

lib/davinci_pas_test_kit/cross_suite/pas_subscription_verification.rb
abhinandan2012: Updated - replaced the meta.profile check with a resource_is_valid? call that validates the Subscription against the PAS Subscription profile using the FHIR validator.

lib/davinci_pas_test_kit/cross_suite/pas_subscription_verification.rb
```diff
@@ -1,32 +1,107 @@
...
+      add_message('error', %(
             The Subscription must include a single filter on the submitting organization
             in the `Subscription.criteria.extension` element.
           ))
+    end
+
+    def verify_pas_subscription_v220(subscription)
+      verify_v220_profile(subscription)
+      verify_v220_rest_hook_channel(subscription)
+      verify_v220_full_resource_payload(subscription)
```
karlnaden: The [profile](https://build.fhir.org/ig/HL7/davinci-pas/en/StructureDefinition-profile-subscription.html) validation takes care of the rest hook channel and the full resource payload, so these don't need separate methods. See duplication for the payload: 

Image

lib/davinci_pas_test_kit/cross_suite/pas_subscription_verification.rb
```diff
@@ -1,32 +1,107 @@
...
+          The Subscription payload content type must be `full-resource` for PAS v2.2.0,
+          but found `#{payload_content || 'none'}`.
+        ))
+    end
+
+    def verify_v220_criteria_format(subscription)
+      filter_criteria = subscription.dig('_criteria', 'extension')
+        &.select { |ext| ext['url'] == BACKPORT_FILTER_CRITERIA_URL }
+
+      if filter_criteria.blank?
```
karlnaden: Looking at [the profile](https://build.fhir.org/ig/HL7/davinci-pas/en/StructureDefinition-profile-subscription.html), I think the validator will check for a single filter criteria - test to verify, but if I'm right, you should remove this error and only have the format part.

lib/davinci_pas_test_kit/cross_suite/pas_subscription_verification.rb
```diff
@@ -1,32 +1,107 @@
...
+        &.select { |ext| ext['url'] == BACKPORT_FILTER_CRITERIA_URL }
+      return if filter_criteria&.length == 1
+
+      add_message('error', %(
             The Subscription must include a single filter on the submitting organization
             in the `Subscription.criteria.extension` element.
           ))
+    end
+
+    def verify_pas_subscription_v220(subscription)
```
karlnaden, line 43: With changes to this method, let's update the [220 preset](https://github.com/inferno-framework/davinci-pas-test-kit/blob/support-v2.2.0/config/presets/pas_server_v220_run_against_pas_client.json.erb#L21) to use a conformant subscription.

karlnaden (changes requested):
I'm getting errors when I run the [2.2.0 suites against each other ](https://github.com/inferno-framework/davinci-pas-test-kit/blob/support-v2.2.0/docs/Running-Suites-Against-Each-Other.md)for the pended workflow, even if I change the Subscription payload type to full-resource.

Looks like the same errors are happening in the base 2.2.0 branch, so let me try to fix those and you can rebase.

lib/davinci_pas_test_kit/cross_suite/pas_subscription_verification.rb
abhinandan2012: removed verify_v220_rest_hook_channel, verify_v220_full_resource_payload, and extract_payload_content. The profile validator handles these.

Screenshot 2026-02-18 at 2 52 29 PM

lib/davinci_pas_test_kit/cross_suite/pas_subscription_verification.rb
abhinandan2012: Removed the 'filter criteria blank' error. The profile validator checks cardinality. Kept only the org-identifier=<identifier> format check.

lib/davinci_pas_test_kit/cross_suite/pas_subscription_verification.rb
abhinandan2012: Updated the v2.2.0 preset to use full-resource payload, org-identifier=8189991234 filter criteria, and added the PAS Subscription profile.

karlnaden:
A few additional things I'm noticing as I run the tests:
1. Looks like we have tests that duplicate much of the same, but probably not all, logic for Subscription Validation in both client and server tests. Can you look at the logic and make a recommendation for how to take away the duplication (compare to the 2.0.1 versions as well)
  - server tests 1.01 and 1.02
  - client tests 8.02 and 8.03 (when using the other authentication suite option)
2. It looks like our notification verification needs some updates, and probably the logic that we use to generate a notification when not provided.  Check tests in client group 9.3.3 (in particular 9.3.3.03 which is expecting a Bundle, but got a ClaimResponse - I think that's a preset or response generation issue) and server group 2.3.2 (particularly 2.3.2.04 which isn't finding the notification to analyze as it should)

config/presets/pas_server_v220_run_against_pas_client.json.erb
```diff
@@ -22,7 +22,7 @@
...
...e Notification Subscription\",\n  \"criteria\" : \"http://hl7.org/fhir/us/davinci-pas/SubscriptionTopic/PASSubscriptionTopic\",\n  \"_criteria\" : {\n    \"extension\" : [{\n      \"url\" : \"http://hl7.org/fhir/uv/subscriptions-backport/StructureDefinition/backport-filter-criteria\",\n      \"valueString\" : \"org-identifier=8189991234\"\n    }]\n  },\n  \"channel\" : {\n    \"extension\" : [{\n      \"url\" : \"http://hl7.org/fhir/uv/subscriptions-backport/StructureDefinition/backport-timeout\",\n      \"valueUnsignedInt\" : 60\n    },\n    {\n      \"url\" : \"http://hl7.org/fhir/uv/subscriptions-backport/StructureDefinition/backport-max-count\",\n      \"valuePositiveInt\" : 20\n    }],\n    \"type\" : \"rest-hook\",\n    \"header\": [\"Authorization: Bearer SAMPLE_TOKEN\"],\n    \"endpoint\" : \"http://localhost:4567/custom/subscriptions_r5_backport_r4_server/subscription/channel/notification_listener\",\n    \"payload\" : \"application/fhir+json\",\n    \"_payload\" : {\n      \"extension\" : [{\n        \"url\" : \"http://hl7.org/fhir/uv/subscriptions-backport/StructureDefinition/backport-payload-content\",\n        \"valueCode\" : \"full-resource\"\n      }]\n    }\n  }\n}"
```
karlnaden: take out meta.profile

abhinandan2012: Updated

lib/davinci_pas_test_kit/client/jobs/send_pas_subscription_notification.rb
```diff
@@ -114,15 +128,15 @@ def send_notification(request_body)
...
+      def persist_notification_request(response, tags, sent_body = nil)
         inferno_request_headers = headers.map { |name, value| { name:, value: } }
         inferno_response_headers = response.headers&.map { |name, value| { name:, value: } }
         requests_repo.create(
           verb: 'POST',
           url: response.env.url.to_s,
           direction: 'outgoing',
           status: response.status,
-          request_body: response.env.request_body,
+          request_body: sent_body || response.env.request_body,
```
karlnaden: is this needed? I would expect the `response.env.request_body` to always have the right content, so not sure why we need to override it in certain cases. Please explain the rationale.

lib/davinci_pas_test_kit/client/response_generator.rb
```diff
@@ -125,42 +129,101 @@ def update_tester_provided_notification_claim_response_entry(notification_bundle
...
+      end
+    end
+
+    def add_v220_notification_entries(mock_notification_bundle, submit_bundle, decision, inner_bundle_full_url)
+      inner_bundle = FHIR::Bundle.new(
+        id: inner_bundle_full_url.delete_prefix('urn:uuid:'),
+        type: 'collection',
+        timestamp: Time.now.utc.iso8601,
+        identifier: FHIR::Identifier.new(system: 'urn:ietf:rfc:3986', value: "urn:uuid:#{SecureRandom.uuid}"),
+        meta: FHIR::Meta.new(profile: ['http://hl7.org/fhir/us/davinci-pas/StructureDefinition/profile-pas-response-bundle'])
```
karlnaden: don't set meta.profile

lib/davinci_pas_test_kit/client/response_generator.rb
```diff
@@ -125,42 +129,101 @@ def update_tester_provided_notification_claim_response_entry(notification_bundle
...
-          update_claim_response_decisions(entry.resource, decision) if entry.resource.resourceType == 'ClaimResponse'
-          entry.request = FHIR::Bundle::Entry::Request.new(
-            method: 'POST',
-            url: "#{fhir_base_url}/#{entry.resource.resourceType}"
-          )
-          entry.response = FHIR::Bundle::Entry::Response.new(
-            status: '201'
-          )
-          mock_notification_bundle.entry << entry
+        add_legacy_notification_entries(mock_notification_bundle, submit_bundle, decision, fhir_base_url)
```
karlnaden: I don't like "legacy" here, can we use the ig version, such as `add_v201_notification_entries`

lib/davinci_pas_test_kit/server/v2.2.0/pas_server_subscription_notification_wait_test.rb
```diff
@@ -7,6 +7,8 @@ class PASSubscriptionNotificationWaitTest < AbstractPASSubscriptionNotificationW
       include ServerURLs
 
       id :pas_server_v220_subscription_notification_wait
+
+      receives_request :notification
```
karlnaden: shouldn't need this. Is there a reason you added it?

lib/davinci_pas_test_kit/server/generated/v2.2.0/pas_server_pended_use_case_group.rb
karlnaden, line 1: You can't add content directly to generated files. Look in the generator:  https://github.com/inferno-framework/davinci-pas-test-kit/blob/support-v2.2.0/lib/davinci_pas_test_kit/generator/templates/use_case_group.rb.erb and https://github.com/inferno-framework/davinci-pas-test-kit/blob/support-v2.2.0/lib/davinci_pas_test_kit/generator/use_case_group_generator.rb

karlnaden (changes requested):

lib/davinci_pas_test_kit/client/response_generator.rb
abhinandan2012: renamed legacy to v201

lib/davinci_pas_test_kit/client/response_generator.rb
abhinandan2012: removed meta.profile from the inner response bundle in add_v220_notification_entries

lib/davinci_pas_test_kit/server/v2.2.0/pas_server_subscription_notification_wait_test.rb
abhinandan2012: removed receives_request :notification from the v2.2.0 notification wait test, since the v2.0.1 versions don't have it and it's unnecessary here.

lib/davinci_pas_test_kit/server/generated/v2.2.0/pas_server_pended_use_case_group.rb
abhinandan2012: moved the v2.2.0 full resource conditional logic into the generator template (use_case_group.rb.erb) and regenerated the files.

lib/davinci_pas_test_kit/client/jobs/send_pas_subscription_notification.rb
abhinandan2012: This override is actually necessary for Inferno's error handling. If Faraday hits a connection error (e.g., tester's endpoint is down), it throws an exception rather than returning a response. We catch that and create a mock Faraday::Response to surface the error in the UI. Because it's a mock response, response.env.request_body is nil. Passing sent_body explicitly ensures the tester can still see the payload we attempted to send alongside the connection error.

karlnaden (changes requested):
More work needed on the Subscription Notification verification tests in the pended workflow. Three things I'm seeing when running the client tests against the server ones
1. failures around SubscriptionStatus entry pointers like: "The Notification Bundle does not include a resource entry for the reference found in SubscriptionStatus.notificationEvent.additional-context with id urn:uuid:0b321e9e-571b-4f83-8cea-fa23424383bb" when generating a notification.
2. inconsistent results between client and server tests
3. Lots of error duplication, e.g., in tests 9.3.3.01-03 - it looks like profile validation is currently being done in all three which is not ideal.

lib/davinci_pas_test_kit/client/endpoints/claim_endpoint.rb
karlnaden: Just noticed this potential bug that is worth fixing:

```suggestion
      content_ext = subscription.dig('channel', '_payload', 'extension')&
```

lib/davinci_pas_test_kit/client/jobs/send_pas_subscription_notification.rb
```diff
@@ -100,9 +103,19 @@ def await_subscription_creation
...
+        # with PAS-specific structure (focus points to an inner PAS Response Bundle).
+        # Using derive_event_notification would flatten this structure, making the focus
+        # point to the ClaimResponse directly instead of the wrapping Bundle.
+        # For v2.0.1, derive_event_notification correctly builds the notification.
+        event_json = if ig_version == 'v2.2.0'
+                       @notification_json
+                     else
+                       derive_event_notification(@notification_json, subscription_full_url, subscription_topic,
+                                                 1).to_json
+                     end
```
karlnaden: This is not correct. The [derive_event_notification method in the subscriptions_test_kit](https://github.com/inferno-framework/subscriptions-test-kit/blob/main/lib/subscriptions_test_kit/suites/subscriptions_r5_backport_r4_client/common/subscription_simulation_utils.rb#L30) updates timestamps and the subscription status details but doesn't do any wrapping or updates to the notification contents. It needs to be applied for either version.

lib/davinci_pas_test_kit/client/jobs/send_pas_subscription_notification.rb
karlnaden: Then add this to the mocked response object (along with headers!) when the error is caught rather than spreading the error handling around:

I think adding this to line 127 would do it -  `request_body:, request_headers: headers`

lib/davinci_pas_test_kit/client/response_generator.rb
karlnaden, line 1: I keep getting confused by the `submit_bundle` variable and whether it is the request or response. It is the response, so let's update it throughout this file to `submit_response_bundle` for clarity.

lib/davinci_pas_test_kit/client/response_generator.rb
```diff
@@ -125,42 +129,106 @@ def update_tester_provided_notification_claim_response_entry(notification_bundle
...
+        response: FHIR::Bundle::Entry::Response.new(status: '201')
+      )
+    end
+
+    def prepare_v220_claim_response(claim_response, decision)
+      update_claim_response_decisions(claim_response, decision)
+      claim_response.outcome = 'complete'
+      return if claim_response.request.present?
+
+      claim_response.request = FHIR::Reference.new(reference: "urn:uuid:#{SecureRandom.uuid}")
```
karlnaden: I don't think we want to update the request if it is not present - should have been populated earlier if possible. Not a good idea to put random values into a reference field.

lib/davinci_pas_test_kit/client/response_generator.rb
```diff
@@ -125,42 +129,106 @@ def update_tester_provided_notification_claim_response_entry(notification_bundle
...
+    end
+
+    def add_v201_notification_entries(mock_notification_bundle, submit_bundle, decision, fhir_base_url)
+      submit_bundle.entry.each do |entry|
+        update_claim_response_decisions(entry.resource, decision) if entry.resource.resourceType == 'ClaimResponse'
+        entry.request = FHIR::Bundle::Entry::Request.new(
+          method: 'POST',
+          url: "#{fhir_base_url}/#{entry.resource.resourceType}"
+        )
+        entry.response = FHIR::Bundle::Entry::Response.new(status: '201')
```
karlnaden: these need to be kept and included when generating 2.2.0 notifications

lib/davinci_pas_test_kit/client/response_generator.rb
```diff
@@ -125,42 +129,106 @@ def update_tester_provided_notification_claim_response_entry(notification_bundle
     end
 
     def build_mock_notification_bundle(notification_timestamp, subscription_reference, subscription_topic,
```
karlnaden, line 143: Plan: update 2.0.1 full-resource support to send the entire response bundle instead of just the claim response
- Rationale: The 2.0.1 PAS IG says full resource can be supported but doesn’t say anything about what the payload looks like. We made a guess that it was the ClaimResponse, but now that we know it is not, it will be better to make them the same and we have the leeway to do so because we said that test kit details could change.
- High-level changes
    - Never need any additional context entries
    * determine_notification_focus - (generated) bundle id if full-resource, claimresponse id otherwise (can’t think of anything better for id-only) - probably need to re-jigger how this method works
    * add_notification_entries - run only if full resource, always add a single entry for the submit_response_bundle - see other comments for additional details on this

lib/davinci_pas_test_kit/cross_suite/pas_notification_verification.rb
karlnaden, line 1: Most of the validation done here is performed already in the subscription test kit. I think you can just verify that there are additional entries beyond the first subscription status and that they contain PAS Response Bundles.

karlnaden (changes requested):
Reviewed the logic and have some requested updates.  Haven't yet looked closely at example runs, but I think with these updates we'll be close to finishing this one.

lib/davinci_pas_test_kit/client/endpoints/claim_endpoint.rb
abhinandan2012: Fixed.

lib/davinci_pas_test_kit/client/jobs/send_pas_subscription_notification.rb
abhinandan2012: Added request_body and request_headers directly to the mocked response.

lib/davinci_pas_test_kit/client/jobs/send_pas_subscription_notification.rb
abhinandan2012: Updated to apply derive_event_notification for both v2.0.1 and v2.2.0.

lib/davinci_pas_test_kit/client/response_generator.rb
abhinandan2012: Renamed submit_bundle to submit_response_bundle throughout the file for clarity.

lib/davinci_pas_test_kit/client/response_generator.rb
abhinandan2012: Removed the logic that was inserting random values into the request reference field.

lib/davinci_pas_test_kit/client/response_generator.rb
abhinandan2012: Updated the logic to ensure we keep all necessary entries when generating v2.2.0 notifications.

lib/davinci_pas_test_kit/client/response_generator.rb
abhinandan2012: Updated the full-resource logic for v2.0.1 to send the entire response bundle (removed the old add_v201_notification_entries mapping method).

lib/davinci_pas_test_kit/cross_suite/pas_notification_verification.rb
abhinandan2012: Simplified the validation to only verify that the additional entries contain the correct PAS Response Bundles.

karlnaden (changes requested):
Code looking good.

To help verify, I'd like to see inputs for two executions of the client vs server tests:
1. passes all relevant tests related to subscriptions
2. has some kind of failure in each relevant test related to subscriptions

Represent them as presets for now (might change at some point). The current v220 client vs server presets will be a good starting point - they fail a lot of the tests.

I think the relevant tests are:
- Client
  - 8.02
  - 8.03 (NOTE: the error "Value is 'requested' but is fixed to 'active' in the profile" will be going away in the published version, so you can leave that in the "passing" execution and don't count it in the "failing" execution)
  - 8.04
  - 9.3.3.01
  - 9.3.3.02
  - 9.3.3.03
  - 9.3.3.04
- Server 
  - 1.01
  - 1.02 (NOTE: the error "Value is 'requested' but is fixed to 'active' in the profile" will be going away in the published version, so you can leave that in the "passing" execution and don't count it in the "failing" execution)
  - 1.03
  - 1.04
  - 1.05
  - 2.3.2.02
  - 2.3.2.03
  - 2.3.2.04

Don't hesitate to reach out if you're having trouble getting a set of inputs that passes or fails a specific test. This is a relatively new ask (particularly the failure execution) so what is possible and helpful is not completely clear yet.

karlnaden (changes requested):
Passing preset looks good. Hoping for a bit more failures on the failure preset. If you aren't able to get errors like I think you can let me know and we can work through it.

- Client
  - 8.02 - can you add a bad status value that isn't allowed, or something like that to get this to fail
  - 8.03 - criteria good, but let's make it full resource to enable other failures later
  - 8.04 - fine to have it pass on second thought - the Subscription test kit generates this and we can't control it.
  - 9.3.3.01 -  break some subscriptions invariant, like make it have type = collection
  - 9.3.3.02 - fine to have this pass - more important to see full-resource notification checks fail
  - 9.3.3.03 - would like to see the error deeper in the Notification Bundle since we're doing a profile validation of it, right?
  - 9.3.3.04 - fine to have it pass on second thought - won't be easy or valuable to get the server test to reject it
- Server
  - 1.01 - same as client 8.02 I think
  - 1.02 - criteria good, but let's make it full resource to enable other failures later
  - 1.03 - fine on second thought - won't be easy or valuable to show failing
  - 1.04 - fine on second thought - won't be easy or valuable to show failing
  - 1.05 - fine on second thought - won't be easy or valuable to show failing
  - 2.3.2.02 - make the notification full-resource and break some notification invariant, e.g., Bundle type = collection
  - 2.3.2.03 - make the notification full-resource and break some full-resource notification invariant, e.g., in the PAS response Bundle itself have a formatting error on the timestamp or something like that.
  - 2.3.2.04 - would like to see the error deeper in the Notification Bundle since we're doing a profile validation of it, right?

execution_scripts/pas_client_v220_run_against_pas_server/pas_v220_client_server_pass.yaml
```diff
@@ -0,0 +1,242 @@
...
+    status: waiting
+    last_completed: 1.06
+    command: "curl -o /dev/null -s '{inferno_base_url}/custom/davinci_pas_client_suite_v220/resume_pass?token={client.session_id}'"
+    action_description: GET the attestation URL
+
+  - state_description: Wait at Client test 1.07 (confirm configuration)
+    session: client
+    status: waiting
+    last_completed: 1.07
+    command: "curl -o /dev/null -s '{inferno_base_url}/custom/davinci_pas_client_suite_v220/resume_pass?token={client.session_id}'"
```
karlnaden: This isn't the right url for this test. To make it easier, I'm moving to have these tests contain outputs that contain the resume_pass url explicitly. Some are in the Subscriptions test kit which is why you have to import the branch of that test kit to get this to work.  Check your curl commands against the ones in my example script, e.g., here https://github.com/inferno-framework/davinci-pas-test-kit/blob/id45-execution-scripts/execution_scripts/pas_client_vs_server_suites/pas_client_vs_server_suites.yaml#L28

karlnaden (changes requested):

execution_scripts/pas_client_v220_run_against_pas_server/pas_v220_client_server_pass.yaml
abhinandan2012: URL corrected.

lib/davinci_pas_test_kit/client/v2.2.0/pas_client_pended_group.rb
karlnaden:
```suggestion
             title: '[USER INPUT VERIFICATION] Inferno's event notification Bundle is conformant',
```

karlnaden:
```suggestion
             title: '[USER INPUT VERIFICATION] Inferno's event notification Bundle matches the Subscription',
```

lib/davinci_pas_test_kit/client/v2.2.0/pas_client_pended_group.rb
```diff
@@ -95,9 +92,7 @@ class PASClientPendedGroup < Inferno::TestGroup
                  notification_bundle: { optional: true } # doesn't use the input (bug in Subscriptions)
                }
              }
-        # test for PAS-specific requirements? Current decision: no, there isn't anything hard in the spec
-        # and testers have to demonstrate and attest that their systems work, which will require some
-        # correspondence.
+        test from: :pas_notification_pas_conformance_test
```
karlnaden: Add a different title for this suite: 

`title: '[USER INPUT VERIFICATION] Inferno's Notification conforms to PAS-specific requirements'`

karlnaden (changes requested):
Looking good.  For 2.2.0, I think that server test 1.01 and client test 8.02 (and corresponding ones for other auth choices) are duplicative of the PAS-specific subscript checks since it now does a profile validation.  Please remove those tests for the 2.2.0 version only.

execution_scripts/pas_client_v220_run_against_pas_server/pas_v220_client_server_pass.yaml
```diff
@@ -0,0 +1,242 @@
...
+    suite_options:
+      client_type: other_auth
+
+  - suite: davinci_pas_server_suite_v220
+    name: server
+    preset: pas-server-v220-run-against-pas-client-passing-tests
+
+normalized_strings:
+  - replacement: <UUID>
+    pattern: /[0-9a-f]{8}-[0-9a-f]{4}-[0-9a-f]{4}-[0-9a-f]{4}-[0-9a-f]{12}/i
```
karlnaden: update to the yaml format

```suggestion
comparison_config:
  normalized_strings:
    - replacement: <UUID>
      pattern: /[0-9a-f]{8}-[0-9a-f]{4}-[0-9a-f]{4}-[0-9a-f]{4}-[0-9a-f]{12}/i
```

execution_scripts/pas_client_v220_run_against_pas_server/pas_v220_client_server_fail.yaml
```diff
@@ -0,0 +1,242 @@
...
+    suite_options:
+      client_type: other_auth
+
+  - suite: davinci_pas_server_suite_v220
+    name: server
+    preset: pas-server-v220-run-against-pas-client-failing-tests
+
+normalized_strings:
+  - replacement: <UUID>
+    pattern: /[0-9a-f]{8}-[0-9a-f]{4}-[0-9a-f]{4}-[0-9a-f]{4}-[0-9a-f]{12}/i
```
karlnaden: update to the yaml format

```suggestion
comparison_config:
  normalized_strings:
    - replacement: <UUID>
      pattern: /[0-9a-f]{8}-[0-9a-f]{4}-[0-9a-f]{4}-[0-9a-f]{4}-[0-9a-f]{12}/i
```

lib/davinci_pas_test_kit/client/v2.2.0/pas_client_pended_group.rb
abhinandan2012: Updated.

abhinandan2012: Updated.

lib/davinci_pas_test_kit/client/v2.2.0/pas_client_pended_group.rb
abhinandan2012: Updated.

execution_scripts/pas_client_v220_run_against_pas_server/pas_v220_client_server_pass.yaml
abhinandan2012: Updated.

execution_scripts/pas_client_v220_run_against_pas_server/pas_v220_client_server_fail.yaml
abhinandan2012: Updated.

lib/davinci_pas_test_kit/server/v2.2.0/pas_server_subscription_input_conformance.rb
```diff
@@ -16,11 +17,30 @@ class PASServerSubscriptionInputConformance < Inferno::Test
...
+
+        updated_subscription = subscription.to_json
+        output(updated_subscription:)
+
+        verify_pas_subscription(updated_subscription, ig_version: 'v2.2.0')
       end
     end
   end
 end
+
```
karlnaden: check rubocop - it doesn't like this line.

lib/davinci_pas_test_kit/server/v2.2.0/pas_server_subscription_setup.rb
```diff
@@ -15,7 +15,7 @@ class PASServerSubscriptionSetup < Inferno::TestGroup
       input_order :server_endpoint, :smart_credentials, :access_token, :subscription_resource
       run_as_group
 
-      test from: :subscriptions_r4_server_subscription_conformance do
+      test from: :pas_server_v220_subscription_input_conformance do
         input :subscription_resource,
               title: 'Pended Prior Authorization Subscription',
```
karlnaden, line 20: add `type: 'textarea'` since no longer inherited from the subscriptions test kit test

karlnaden (changes requested):
two minor fixes, then I'll approve and merge

lib/davinci_pas_test_kit/server/v2.2.0/pas_server_subscription_setup.rb
abhinandan2012: Added.

lib/davinci_pas_test_kit/server/v2.2.0/pas_server_subscription_input_conformance.rb
abhinandan2012: Fixed.

karlnaden (approved):

PR #59, opened 2026-03-27
Add simulation flags to groups and tests

ljtucker:
# Summary

# Testing Guidance


arunkumarhk (approved):

lib/davinci_pas_test_kit/generator/validation_test_generator.rb
```diff
@@ -109,14 +111,19 @@ def generate
...
       def user_input?
         request_type.include?('request')
       end
 
+      # rubocop:disable Naming/PredicateMethod
+      def simulation_verification
+        user_input?
+      end
+      # rubocop:enable Naming/PredicateMethod
+
```
karlnaden: Simple refactor to follow the style rather than disabling it, could be done for `skip_if_empty` in several places as well.

```suggestion
      def simulation_verification?
        user_input?
      end
```

lib/davinci_pas_test_kit/generator/templates/group.rb.erb
```diff
@@ -112,9 +112,24 @@ module DaVinciPASTestKit
           - [PAS PractitionerRole](https://hl7.org/fhir/us/davinci-pas/STU2/StructureDefinition-profile-practitionerrole.html)
         )<% end %>
         run_as_group
-        <% subgroup.each do |category, id_list| %>
-        group do
-          title '<%= category%>'
+        <%- subgroup.each do |category, id_list| -%>
+<%-
+  title_lines = category.to_s.split("\n").map(&:strip)
```
karlnaden, line 117: I'm not sure why this got so much more complicated, but it doesn't look like it is using the erb templating paradigm. Can you explain the intention of this update so we can plan how to implement it in a cleaner way?

lib/davinci_pas_test_kit/generator/group_generator.rb
```diff
@@ -219,19 +221,27 @@ def grouped_must_support_test_ids
...
+        inquiry_request_key = [
+          'Inquiry Request Must Support',
+          "#{' ' * 10}verifies_requirements 'hl7.fhir.us.davinci-pas_2.0.1@36'",
+          "#{' ' * 10}simulation_verification"
+        ].join("\n")
+
         {
           '$submit Element Support' => {
             'Submission of claims to the $submit operation for must support validation' => submit_tests,
-            "[USER INPUT VALIDATION] Submit Request Must Support'" \
```
karlnaden, line 225: Looks like this is the source of the complication in the `group.rb.erb` file. Generators harbor a lot of nutty code in the name of complex code re-use. I guess we leave it.

karlnaden (changes requested):
updating rubocop is dangerous partly because of these new refactors, which seem to often be more trouble than they are worth. Probably a good idea to decide our policy on this kind of thing so that we have a go to approach. I'm fine with most of the changes as is, but there are some low-hanging fruit that I think better to actually address rather than disable.

I'm also inclined to have this merged into the `support-v2.2.0` branch first instead of straight to main since that is where most of our activity is going.

Also - gemspec needs to be updated to have inferno_core 1.1.2 as the minimum required version:

`spec.add_runtime_dependency 'inferno_core', '~> 1.1', '>= 1.1.2'`

karlnaden:
moved into the 2.2 branch instead

PR #64, opened 2026-04-17
ID-55 Bundle Validation Patch

ljtucker:
### Problem
When validating a PAS Bundle against its profile, the FHIR validator emits issues for both the bundle as a whole and for each entry resource individually. Because the test kit already validates each entry resource standalone (against its own PAS/US Core profile), the bundle-level entry messages were duplicated — appearing once from the bundle pass and again from the individual resource pass. This produced noisy, confusing test output.

### Changes

**`pas_bundle_validation.rb`**
- Extracted bundle validation into a new `bundle_profile_valid?` method that calls `resource_is_valid?` with `add_messages_to_runnable: false`, then filters the raw issues before deciding pass/fail
- Added `reject_entry_resource_issues` to strip issues located at `Bundle.entry[N].resource` (and any pre-filtered issues), leaving only true bundle-structural issues to surface
- Added a `TODO (ID-61)` comment noting that resources with no PAS profile (e.g. `DocumentReference`, `Encounter`) are currently silently dropped by this filter and not individually validated until that ticket is resolved

**`pas_bundle_validation_spec.rb`**
- Added a `describe '#reject_entry_resource_issues'` block with three examples:
  - Keeps bundle-structural issues (slicing hints at `Bundle.entry[N]`, bundle-level errors) and rejects entry-resource issues
  - Rejects issues pre-marked as `filtered` regardless of location
  - Returns an empty array for empty input

### Validation
All 8 affected tests (4 server + 4 client, across v2.0.1 and v2.2.0) were run before and after. Entry-resource duplicate messages are gone; all standalone resource messages are preserved. No unexpected losses.


karlnaden (changes requested):
Code looks good. One primary goal with this task is to get to the point where we can have example PAS bundles that actually pass validation. Can you take a crack at updating the presets to include something that fully passes? If you run into issues that seem outside the scope of this ticket, no need to fix, but create a JIRA ticket so that we can address.

karlnaden (approved):

PR #65, opened 2026-05-05
ID-61: Add US Core fallback profile inference for v2.2.0 PAS bundle validation

abhinandan2012:
## Summary

Implements ID-61: adds US Core 6.1.0 profile fallback for PAS v2.2 Bundle validation.

Bundle entries that receive no PAS profile via reference-based traversal are now validated against an applicable US Core 6.1.0 profile if one exists, or against the base FHIR R4 definition if no US Core profile applies. Fallback is limited to PAS v2.2.x only.

---

### Core Logic (`pas_bundle_validation.rb`)

- Adds US Core 6.1.0 fallback profile inference for Bundle entries that receive no PAS profile from reference traversal.
- Profile assignment follows this priority order:
  1. PAS reference-based traversal (existing)
  2. Declared `meta.profile` values for entries not reached by traversal (`add_declared_profiles_to_unprofiled_entries`)
  3. US Core 6.1.0 fallback (`add_us_core_profiles_to_unprofiled_entries`)
  4. Base FHIR R4 definition for entries with no applicable US Core profile (`:base_r4` sentinel, validated via `resource_is_valid?` with no profile URL)
- Fixes broken traversal of `Claim.extension:encounter` — the PAS Encounter reference uses a cross-version FHIR R5 extension not captured by generated metadata. Added manual FHIRPath traversal in `handle_claim_profile` to correctly assign `profile-encounter` to Encounter entries. Verified working via debug output confirming the path returns a `Reference` type to `Encounter/SurgicalEncounterExample`.
- Supports multi-profile US Core resources:
  - **Condition**: category-based (`encounter-diagnosis` vs `problems-health-concerns`). Falls back to `problems-health-concerns` when no known category is found.
  - **DiagnosticReport**: `LAB` category → lab profile, else note profile.
  - **Observation**: LOINC code-specific profiles first (17 mappings from US Core 6.1.0 vital signs and specialty profiles), then category-based fallback (`social-history`, `vital-signs`, `survey`, `laboratory`), then `simple-observation` as final catch-all.
- Limits fallback to PAS `2.2.x` via `us_core_profile_fallback_enabled?`. v2.0.1 bundles are completely unaffected.
- Fixes `metadata_map` to cache results per-version, preventing stale metadata when validating across multiple IG versions in the same test run.
- Adds `reset_bundle_profile_inference_state` to clear profile map and bundle entry map state between bundle validations, preventing stale data from a previous bundle affecting the next.
- Fixes `reference_elements.dup` in `extract_profiles_to_validate_each_entry` to prevent mutation of the shared metadata reference array across calls.
- Adds `profile_url_without_version` helper and uses it throughout metadata lookups so that versioned profile URLs (e.g. `url|6.1.0`) are correctly resolved against the metadata map.
- Adds `profile_url_for_validation` to correctly append version per IG:
  - US Core profiles → `|6.1.0`
  - PAS profiles → `|{pas_version}`
  - URLs already containing `|` or matching base profile → used as-is

### Validator Configuration

- Adds `hl7.fhir.us.core#6.1.0` to validator IG list for v2.2 client suite (`client/v2.2.0/client_suite.rb`) so the validator knows US Core profiles when validating Bundle entries.
- Adds `hl7.fhir.us.core#6.1.0` to v2.2 server suite via `server_suite_generator.rb` and `server_suite.rb.erb` template, using a new `validator_igs_expression` method that conditionally appends the US Core IG for v2.2.x only.

### Test Fixtures (`pas_server_v220_run_against_pas_client.json.erb`)

- Adds a `supportingInfo` entry with custom category `urn:inferno:pas-codes/default-profile-testing` (sequence 99) to all v2.2 submit request preset bundles (approval, denial, pended, must support). This category code falls outside all PAS profile slices so it is never reached by PAS reference traversal, demonstrating the fallback code path is exercised.

---

## Known Gaps / Follow-up

- **Base R4 fallback fixture**: The `:base_r4` fallback is implemented in code but no preset resource currently exercises it — every resource in the presets either has a PAS profile or a US Core profile. A follow-up should add a fixture entry for a resource type with no US Core profile (e.g. `Endpoint`) to verify this path end-to-end.

- **True US Core fallback fixture**: The `default-profile-testing` entry (sequence 99) references `DocumentReference/episode-summary`, which is already reached by PAS traversal via sequence 1 (`additionalInformation`). So the entry demonstrates the code path exists but does not fully isolate US Core fallback inference. A dedicated resource reachable only via the custom category would be needed to demonstrate true fallback in isolation.

- **Encounter extension traversal is a temporary workaround**: The `Claim.extension:encounter` FHIRPath fix in `handle_claim_profile` is a manual patch for a gap in the generated metadata. A follow-up task will investigate replacing the current traversal handling with shared metadata-driven navigation logic so extension references can be resolved without manual workarounds.

---

## Unit Testing

```sh
bundle exec rspec
bundle exec rubocop
```

---

## UI Testing Steps

Use the v2.2 client and server suites run against each other locally. See [Running Suites Against Each Other](https://github.com/inferno-framework/davinci-pas-test-kit/blob/support-v2.2.0/docs/Running-Suites-Against-Each-Other.md) for setup details.

### Setup

1. Start Inferno and open the UI.
2. Create two sessions:
   - **Client Suite**: `Da Vinci PAS Client Suite v2.2.0` (`davinci_pas_client_suite_v220`)
   - **Server Suite**: `Da Vinci PAS Server Suite v2.2.0` (`davinci_pas_server_suite_v220`)

### Run Flow

1. In the **client session**, click **Run All Tests** and proceed 
   through prompts until it waits for server interaction.
2. In the **server session**, run test groups in order:
   - Subscription Setup
   - Approval Workflow
   - Denial Workflow
   - Pended Workflow
   - Must Support and Error Handling

### What To Verify

Focus on Bundle validation result rows in both suites.

**Validator configuration:**
- Validator loads both `hl7.fhir.us.davinci-pas#2.2.0` and `hl7.fhir.us.core#6.1.0`.

**PAS profile precedence (should NOT fall back to US Core):**
- `Encounter/SurgicalEncounterExample` → validates against `profile-encounter` via `Claim.extension:encounter` FHIRPath fix. No `6.1.0` reference should appear for this resource.
- `ServiceRequest/ReferralRequestExample` → validates against `profile-servicerequest` via `requestedService` extension traversal.
- `Patient/SubscriberExample` → validates against `profile-subscriber`.

**US Core via declared meta.profile:**
- `DocumentReference/episode-summary` → validates against `us-core-documentreference|6.1.0` because it declares that profile in `meta.profile`, honoured during PAS traversal.

**General:**
- Validation succeeds if the resource conforms to at least one applicable target profile.
- No resource that has a PAS profile should fall through to US Core.

### Regression Check

Run the **PAS v2.0.1 suite** and confirm US Core fallback is **not** applied — no `us-core-*` profiles or `6.1.0` references should appear in any validation messages.

---

## Bundle Entries Using US Core Profiles

The predefined v2.2.0 PAS presets primarily exercise PAS reference-based profile inference. The following notes explain what each entry actually demonstrates.

### Declared US Core Profile via PAS Traversal

`DocumentReference/episode-summary` validates against US Core 6.1.0 because it declares a US Core profile in `meta.profile`, which is honoured by `add_declared_profiles` during normal PAS reference traversal (reached via `supportingInfo:additionalInformation`, sequence 1).

- Declared profile: `us-core-documentreference`
- Validator target: `us-core-documentreference|6.1.0`

### Custom Category Entry (sequence 99)

A `supportingInfo` entry with category `urn:inferno:pas-codes/default-profile-testing` was added to all submit request presets. This category falls outside all PAS profile slices so it is never reached by PAS traversal demonstrating the fallback code path is reachable.

In the current presets, the referenced `DocumentReference/episode-summary` already has profiles assigned from PAS traversal via sequence 1, so the fallback does not additionally fire for it in isolation. A dedicated resource reachable only via the custom category would be needed to fully demonstrate true US Core fallback inference in isolation (tracked in Known Gaps above).

### PAS Profile Precedence

- `ServiceRequest/ReferralRequestExample` → `profile-servicerequest` (via `Claim.item.extension:requestedService` traversal)
- `Encounter/SurgicalEncounterExample` → `profile-encounter` (via `Claim.extension:encounter` FHIRPath fix in `handle_claim_profile`)


lib/davinci_pas_test_kit/cross_suite/pas_bundle_validation.rb
```diff
@@ -246,6 +319,146 @@ def validate_bundle_entries_against_profiles(version)
...
+        profile_urls = us_core_profile_urls_for_resource(resource)
+        profile_urls = [BASE_R4_PROFILE] if profile_urls.blank?
+
+        profile_urls.each do |profile_url|
+          add_resource_target_profile_to_map(resource_full_url, resource, profile_url)
+        end
+      end
+    end
+
+    def add_declared_profiles_to_unprofiled_entries(bundle_entry, version)
```
karlnaden: I think I see what this method is doing
- prioritize declared profiles in meta.profile before defaulting to us core / base resource profiles
- add additional profile constraints on referenced instances based on declared meta.profile to catch references not discovered during the primary reference walk to further avoid defaulting

However, I don't think this is the right approach for two reasons:
1. declared profiles are system-added constraints that don't come from the PAS IG. The validator will use those declared profiles in its validation, but we don't need to validate the references as well because it isn't part of the IGs requirements.
2. The concern about "missed" references is warranted, but misplaced. Based on a review of PAS profiles, all references in PAS profiles that constrain profiles are marked as must support meaning that they should be captured in the extracted metadata and checked for references during the primary reference walk. Thus, I think our efforts should be focused on making sure that the metadata extraction and reference walk are complete is how to handle this potential gap.

Based on that, my recommendation is to remove this method and stick with the two step process: primary reference walk using metadata followed by defaulting to us core / base fhir resource definitions for entries not reached by the walk. We'll work on resolving any issues in the primary reference pass separately.

karlnaden (changes requested):

lib/davinci_pas_test_kit/cross_suite/pas_bundle_validation.rb
abhinandan2012: Fixed by removing the declared meta.profile pass for unreached bundle entries. The flow is now PAS metadata traversal first, then US Core fallback, then base R4 fallback only for entries not reached by that traversal.

lib/davinci_pas_test_kit/cross_suite/pas_bundle_validation.rb
```diff
@@ -246,6 +318,129 @@ def validate_bundle_entries_against_profiles(version)
...
+
+    def category_code_present?(resource, code, system:)
+      codeable_concept_has_code?(resource.category, code, system:)
+    end
+
+    def codeable_concept_has_code?(codeable_concepts, code, system: nil)
+      systems = Array(system).compact
+      Array(codeable_concepts).compact.any? do |codeable_concept|
+        Array(codeable_concept&.coding).compact.any? do |coding|
+          coding.code == code && (systems.blank? || coding.system.blank? || systems.include?(coding.system))
```
karlnaden: What is the purpose behind this permissiveness around the system? It looks like all calls to this method include a specific system. Shouldn't the method require one to be provided and require the system to be present? Is there a case you think that is the wrong semantics?

karlnaden (changes requested):

lib/davinci_pas_test_kit/cross_suite/pas_bundle_validation.rb
abhinandan2012: The permissive check was meant to avoid missing category/code matches when codings omit `system`, but looking through the current callers, I agree stricter semantics are more appropriate since they all pass explicit systems. I’ve updated this to require `coding.system` to match the provided system.

karlnaden (approved):

PR #67, opened 2026-05-08
Happy pass of v2.2.0

ljtucker:
## Summary

Fixes must-support test failures in the v2.2.0 (and v2.0.1) server and client suites.

### Infrastructure
- **New `PASMustSupportMetadataExtractor` override** — handles non-MS parent slices with MS children (`optional_slices` mechanism), fixes `fixedBoolean: false` discriminator matching, normalises FHIRPath discriminator paths, and adds a `save_pattern_slice` override for `patternIdentifier.type.coding` slices (Organization PI identifier)
- **`pas_bundle_validation.rb`** — fixes `find_profile_url` comparison bug and adds v2.2.0 `meta.profile`-aware profile selection for submit Claim entries
- **`validator_suppressions.rb`** — 7 new suppressions for IG context declaration defects (see IG Discrepancies section in attached doc)
- Regenerated all metadata YAMLs; updated `must_support_data_gathering.rb` and `must_support_test.rb` to use the new optional-slices path

### Server Suite — v2.2.0
Preset fixes for tests 3.1.2.01–3.1.2.11, 3.2.1.01, and all workflow payloads: NPI corrections, PractitionerRole telecom, subscription resource updates, Coverage MS elements, Encounter MS elements, Organization PI identifier, Claim.related (must-support preset), and a heavily corrected inquire request bundle.

### Client Suite — v2.2.0
- `response_generator.rb` — new `build_mock_communication_request`, `build_mock_practitioner`, `build_mock_practitioner_role`, `build_mock_task` methods; ClaimResponse extended with ~20 missing v2.2.0 MS elements
- Preset: `ClaimResponse.outcome` `queued` → `complete` for pended response

### Still Failing (known, unresolvable)
| Tests | Reason |
|---|---|
| 3.1.2.04 / 10.2.04 and 3.2.2.03 / 10.6.03 | inferno_core `find_required_binding_slice` bug — unexpandable X12 valueset produces `values: []`; Coverage slice never matched |
| 3.1.3.02 / 10.4.02 and 3.2.3.02 / 10.7.02 | IG discrepancies #6–#8: `extension:authorizedProvider`, `addItem` extensions, `reviewActionCode` have context declarations that contradict their must-support placement in the ClaimResponse profile |

### v2.0.1
Preset corrections for submit/inquire Claim Update, Coverage, and Encounter tests; `response_generator.rb` parity fixes.

## Detailed Documentation
You can read a full breakdown of the changes in this PDF: [pas-get-passing-changes.pdf](https://github.com/user-attachments/files/28023186/pas-get-passing-changes.pdf)


lib/davinci_pas_test_kit/client/response_generator.rb
karlnaden, line 1: In general, I'd prefer to demonstrate must support coverage via mocked responses in a preset rather than making the mocked responses returned by Inferno more complicated. See some context for the response mocking [here in the wiki](https://github.com/inferno-framework/davinci-pas-test-kit/wiki/Client-Details#responses): the upshot is that the mocked stuff should be simple, conformant, based very closely on PAS IG examples, and not expected to cover all must supports.

My primary worry with complicating the mocked examples with lots of extra stuff is making them increasingly non-sensical and so not usable. In contrast, a non-sensical preset that we use for internal testing of the suites against each other feels to me less problematic.

Do you think any of the changes you made in here meet the criteria of "simple, conformant, and based very closely on PAS IG examples". If not, my recommendation would be to update the presets to return the more complex responses to cover must support elements and keep the mocking stuff as it was.

lib/davinci_pas_test_kit/cross_suite/generated/v2.2.0/claim_update/metadata.yml
```diff
@@ -323,6 +323,21 @@
...
+    :discriminator:
+      :type: type
+      :code: Period
+  - :slice_id: Claim.supportingInfo:AdmissionDates.timing[x]:timingPeriod
+    :slice_name: timingPeriod
+    :path: supportingInfo.timing[x]
+    :optional: true
+    :discriminator:
+      :type: type
+      :code: Period
```
karlnaden: Why are these considered optional_slices? Looks like maybe they are an oddity in the [2.2.1 profiles](https://hl7.org/fhir/us/davinci-pas/2.2.1/en/StructureDefinition-profile-claim-inquiry.html)? Seems like the "optional_slices" stuff is only really used for v2.0.1, do I have that right?

lib/davinci_pas_test_kit/cross_suite/must_support/must_support_test.rb
karlnaden, line 1: Did you mean to remove the X12 and data absent reason special case handling that abhi added?

lib/davinci_pas_test_kit/cross_suite/generated/v2.0.1/claim_inquiry/metadata.yml
```diff
@@ -87,47 +87,7 @@
...
     :url: http://hl7.org/fhir/us/davinci-pas/StructureDefinition/extension-reviewActionCode
     :modifier_extension: false
-  :slices:
-  - :slice_id: Claim.careTeam:OverallClaimMember
-    :slice_name: OverallClaimMember
-    :path: careTeam
-    :discriminator:
-      :type: value
-      :values:
-      - :path: extension.where(url='http://hl7.org/fhir/us/davinci-pas/StructureDefinition/extension-careTeamClaimScope').valueBoolean
```
karlnaden, line 97: I don't want to dig too deep into v2.0.1 issues at this time since it does not need to be our current focus, but I noticed that some of the slices aren't getting transferred over to the `optional_slices` list such as this "Claim.careTeam:OverallClaimMember" and that "Claim.careTeam:ItemClaimMember". Does that mean that this [must support extensions](https://github.com/inferno-framework/davinci-pas-test-kit/blob/d4768351661df7041f6cc3bcd17bf7786e84a371/lib/davinci_pas_test_kit/cross_suite/generated/v2.0.1/claim_inquiry/metadata.yml#L46C10-L46C72) at "Claim.careTeam:OverallClaimMember.extension:careTeamClaimScope" and "Claim.careTeam:ItemClaimMember.extension:careTeamClaimScope" aren't going to be able to checked because they are missing navigation information in the `optional_slices` list? Or is there another mechanism for supporting that navigation?

karlnaden (changes requested):
A few initial questions and reactions.

lib/davinci_pas_test_kit/cross_suite/generated/v2.2.0/claim_inquiry/metadata.yml
```diff
@@ -38,87 +38,87 @@
     :modifier_extension: false
   - :id: Claim.extension:levelOfServiceType
     :path: extension
-    :url: http://hl7.org/fhir/us/davinci-pas/StructureDefinition/extension-levelOfServiceCode
+    :url: http://hl7.org/fhir/us/davinci-pas/StructureDefinition/extension-levelOfServiceCode|2.2.0
```
karlnaden: Some pas extension canonical URLs now have a version added to them while some do not. Questions:
- What is causing the version to be specified. Is it from something you added explicitly, or is it coming from core code updates?.
- Is the version needed for some check we're doing? I think that Abhi recently updated navigation to [ignore the version part of the canonical](https://github.com/abhinandan2012/inferno-core/blob/main/lib/inferno/dsl/must_support_assessment.rb#L341), so it would need to be something other than the navigation hat needs it.
- it looks like the version isn't getting specified when there are not multiple versions, but that's hard to tell, so if we're going to keep it there, it should always be there.

lib/davinci_pas_test_kit/cross_suite/pas_bundle_validation.rb
```diff
@@ -197,6 +197,19 @@ def validate_resources_conformance_against_profile(bundle, profile_url, version,
...
+        # (workflow bundles use profile-claim; must-support bundles use profile-claim-update).
+        if version == '2.2.0' && %w[submit submit_request].include?(request_type) &&
+           root_entry.resource.resourceType == 'Claim'
+          meta_profiles = Array(root_entry.resource.meta&.profile)
+          base_claim_url = 'http://hl7.org/fhir/us/davinci-pas/StructureDefinition/profile-claim'
+          if meta_profiles.include?(base_claim_url) && !meta_profiles.include?(root_resource_profile_url)
+            root_resource_profile_url = base_claim_url
+          end
+        end
+
```
karlnaden: The validator will perform any validation against (loaded) profiles indicated in meta.profile. However, we do want to update our code to choose the correct profile that we specify. Let's do that by looking at the related element - it is required for `profile-claim-update` and disallowed fro `profile-claim`. In the future, the test being run may indicate which type it should be, but let's start here.

lib/davinci_pas_test_kit/cross_suite/validator_suppressions.rb
```diff
@@ -140,7 +140,23 @@ module DaVinciPASTestKit
...
+      'extension-productOrServiceCodeEnd is not allowed to be used at this point',
+      # IG defect: extension-infoChanged and modifierextension-infoCancelled context only lists Claim.item,
+      # but profile-claim-base marks them must-support on Claim.supportingInfo as well
+      'extension-infoChanged is not allowed to be used at this point',
+      'modifierextension-infoCancelled is not allowed to be used at this point',
+      # The IG's own PASClaimInquiryExample uses HCPCS code B4184. When the validator checks against
+      # a specific published version of the HCPCS code system (e.g. 2025-01), it reports the code as
+      # unknown because that version snapshot may not include every HCPCS code. Suppressed: the code
+      # is valid per the IG and the code system version mismatch is a validator environment issue.
+      %r{Unknown code .+ in the CodeSystem 'http://www\.cms\.gov/Medicare/Coding/HCPCSReleaseCodeSets'}
```
karlnaden: I think we should resolve this error rather than suppressing it. Let's create a ticket and do so elsewhere

lib/davinci_pas_test_kit/generator/pas_must_support_metadata_extractor.rb
```diff
@@ -1,6 +1,250 @@
...
+  #  that simplifies checking for MustSupport elements.
+  # MustSupport elements may be either plain elements, extensions, or slices.
+  # This logic was originally developed for the US Core Test Kit and has been migrated into Inferno core.
   class PASMustSupportMetadataExtractor < Inferno::DSL::MustSupportMetadataExtractor
+    # Override save_pattern_slice to handle patternIdentifier slices that discriminate on
+    # identifier.type.coding (e.g. Organization.identifier:PI) rather than identifier.system.
+    # inferno_core only extracts .system from patternIdentifier, leaving it nil for type-coded
+    # slices, which causes find_pattern_identifier_slice to never match at runtime.
+    # When patternIdentifier.type.coding is present, emit a patternCodeableConcept discriminator
+    # on path 'type' instead, which inferno_core CAN navigate at runtime.
```
karlnaden, line 14: Is there a more general fix for this that could be added to the core handling? Or is this truly a special case for PAS?

karlnaden (changes requested):

lib/davinci_pas_test_kit/cross_suite/generated/v2.0.1/claim_inquiry/metadata.yml
ljtucker: https://acrosystemsinc.atlassian.net/browse/ID-140 I have added this Jira ticket to the backlog so we can wrap up happy pass work for v2.0.1 in a separate PR

ljtucker: although you were right to note this particular deletion as a regression -- this was meant to be moved to an optional_slice but was instead fully dropped. Corrected this

lib/davinci_pas_test_kit/cross_suite/must_support/must_support_test.rb
ljtucker: That was not supposed to be removed... I have reverted this deletion

lib/davinci_pas_test_kit/cross_suite/generated/v2.2.0/claim_update/metadata.yml
ljtucker: Optional_slices is necessary here due to a limitation in inferno_core. We will get this issue resolved there, and once done we can roll back optional_slices implementation in this kit

lib/davinci_pas_test_kit/cross_suite/generated/v2.2.0/claim_inquiry/metadata.yml
ljtucker: Root cause: a stale override in our PASMustSupportMetadataExtractor.

The `must_support_extensions` and `must_support_extension_elements` methods were overridden in our extractor. That override was written to match inferno_core 1.1.x, which didn't strip version suffixes from extension profile URLs. When the project bumped to inferno_core 1.2.1, the base class gained canonical_url_without_version (which does `url.split('|').first` — but our override, having no super call, kept reading `element.type.first.profile.first` directly, bypassing that stripping.

Solution is to remove the custom override classes and use the ones directly from inferno core now that Abhi's changes handle everything our overrides did and more.

lib/davinci_pas_test_kit/cross_suite/validator_suppressions.rb
ljtucker: Ticket written and suppression reverted: https://acrosystemsinc.atlassian.net/browse/ID-141

lib/davinci_pas_test_kit/generator/pas_must_support_metadata_extractor.rb
ljtucker: I agree, the following functions would benefit more kits than just PAS, so they would be worth migrating to inferno_core:

- save_pattern_slice
  - What to extract: The patternIdentifier.type.coding handling logic (lines 13-25)
  - How: Merge this logic into inferno_core's existing save_pattern_slice method
- take_discriminator_step
  - What to extract: The elsif step_path.include?(' as ') branch (lines 67-76) that handles FHIRPath cast syntax
  - How: Add this branch to inferno_core's discriminator path navigation (if take_discriminator_step exists there) or create it if it doesn't exist
- find_element_by_discriminator_path
  - What to extract: This entire helper method (lines 42-51)
  - How: Check if inferno_core already has this; if not, add it as a general utility; if yes, ensure it uses the enhanced take_discriminator_step
- value_slices
  - This override does not need to be migrated, but will just become redundant once the above are migrated, so it can be deleted from PAS at the conclusion of this work

I do not want the merge conflicts to pile up on this PR while we work on inferno_core, so I have spun up a tech debt ticket to address this once this PR is merged: https://acrosystemsinc.atlassian.net/browse/ID-144

lib/davinci_pas_test_kit/client/response_generator.rb
ljtucker: A few of the changes to the generator I do think were worth adding as they provided useful and sensible additions to the responses, but a large chunk were more complicated than the generator should be producing, just to prove Must Support compliance. I have removed a chunk and instead set up a preset response for the $inquire Must Support group.

lib/davinci_pas_test_kit/generator/must_support_test_generator.rb
karlnaden, line 1: Changes to this file are related to `optional_slices`, so should be rolled back.

lib/davinci_pas_test_kit/cross_suite/pas_bundle_validation.rb
```diff
@@ -665,15 +678,39 @@ def extract_base_url(absolute_url)
...
+    # @param version [String] The IG version (e.g. '2.2.0').
+    # @param claim [FHIR::Claim] The Claim resource to inspect.
+    # @return [String] The profile URL to validate against.
+    def determine_claim_submit_profile_url(version, claim)
+      return PASConstants::CLAIM_PROFILE unless version == '2.2.0'
+
+      if claim.related.blank?
+        'http://hl7.org/fhir/us/davinci-pas/StructureDefinition/profile-claim'
+      else
+        PASConstants::CLAIM_PROFILE
```
karlnaden: extract to a constant for use in all locations referencing. The naming of the `CLAIM_PROFILE` constant is unfortunate (should be `CLAIM_UPDATE_PROFILE`), but I don't think we want to do that refactor here, so maybe `CLAIM_PROFILE_FIRST_SUBMIT`? It can be updated when we refactor the other one.

lib/davinci_pas_test_kit/client/response_generator.rb
```diff
@@ -249,7 +249,49 @@ def build_mock_claim_response(claim, request_bundle, root_url, operation, decisi
...
+        # v2.2.0 removed 'queued' from ClaimResponseOutcome ValueSet; pended status is
+        # conveyed via item.extension:reviewActionCode instead.
+        outcome: decision == :pended && ig_version != 'v2.2.0' ? 'queued' : 'complete',
+        preAuthPeriod: FHIR::Period.new(
+          start: timestamp.strftime('%Y-%m-%d'),
+          end: (timestamp + 1.month).strftime('%Y-%m-%d')
+        ),
+        communicationRequest: [
+          FHIR::Reference.new(display: 'Mock Communication Request')
+        ],
```
karlnaden: I don't think we want to have a display-only reference in the generated data and I don't think we want to create a full CommunicationRequest, so remove.

lib/davinci_pas_test_kit/client/response_generator.rb
```diff
@@ -249,7 +249,49 @@ def build_mock_claim_response(claim, request_bundle, root_url, operation, decisi
...
+        error: [
```
karlnaden: I don't think we want errors in the basic response that Inferno generates. Remove.

lib/davinci_pas_test_kit/client/response_generator.rb
```diff
@@ -315,9 +357,84 @@ def build_mock_bundle(claim_response, request_bundle, root_url, operation, times
...
       response_bundle.entry.concat(referenced_entities(claim_response, request_bundle.entry, root_url))
+
+      if ig_version == 'v2.2.0'
+        practitioner = build_mock_practitioner
+        pract_full_url = "urn:uuid:#{practitioner.id}"
+        response_bundle.entry << FHIR::Bundle::Entry.new(fullUrl: pract_full_url, resource: practitioner)
+
+        practitioner_role = build_mock_practitioner_role(pract_full_url)
+        pract_role_full_url = "urn:uuid:#{practitioner_role.id}"
+        response_bundle.entry << FHIR::Bundle::Entry.new(fullUrl: pract_role_full_url, resource: practitioner_role)
```
karlnaden: Is this PractitionerRole referenced anywhere? I don't see its id being added anywhere. It should be referenced if it is going to be included in the mocked response.

lib/davinci_pas_test_kit/generator/pas_must_support_metadata_extractor.rb
karlnaden: good plan

karlnaden (changes requested):

karlnaden (approved):
With two small changes, I was able to get a complete pass:
- extension "wrong location" errors had a version on my run (maybe related to the updated validator version?) so I updated the message suppression to allow that
- There was a terminology error on a deprecated HCPCS code in the presets that I replaced with a current code

I was a little surprised that we had a complete pass without some of the proposed changes to inferno-core - did those end up only being needed for 2.0.1, or are there 2.2.1 inputs that would trigger some known failures? Either way, I think we have the tickets to fix the additional issues, so this one is good to go assuming you're ok with my updates.

PR #68, opened 2026-05-20
ID 61: Replace FHIRPath-based bundle reference traversal with core fhir_navigation handling

abhinandan2012:
## Summary

Replaces the FHIRPath-based bundle reference traversal in PAS bundle validation with the shared `FHIRResourceNavigation` module from Inferno core.

This refactor removes the dependency on `evaluate_fhirpath` (an external HTTP call to a FHIRPath server) and switches the traversal flow to metadata-driven in-process navigation using `resolve_path`. Extension slice references are now resolved automatically through extracted profile metadata, removing the need for manual FHIRPath workarounds.

---

### Navigation Module (`pas_bundle_validation.rb`)

- Adds `ReferenceNavigationContext` wrapper class that includes `Inferno::DSL::FHIRResourceNavigation` with `metadata` accessible.

- Replaces `evaluate_fhirpath` in `process_reference_element` with `reference_values_for_element` which uses `resolve_path` from the navigation module. The new approach:
  - Runs in-process (no external HTTP call)
  - Returns `FHIR::Reference` objects directly instead of JSON hashes
  - Handles `value[x]` choice types natively
  - Resolves extension slice paths automatically via metadata

- Adds `profile_metadata` parameter to `process_reference_element` so the navigation context has access to `must_supports[:extensions]` for automatic extension URL resolution.

- Adds supporting navigation methods:
  - `reference_values_for_element` — resolves references using the navigation module, filters to `FHIR::Reference` objects with a present reference, and returns their reference strings.
  - `navigation_path_for_reference` — orchestrates path conversion from metadata format to navigation-compatible format.
  - `resource_relative_reference_path` — strips resource type prefix from paths (e.g. `Claim.patient` → `patient`) since `resolve_path` starts from the resource itself.
  - `navigation_compatible_reference_path` — processes each path segment and converts extension slice notation to `where(url='...')` syntax.
  - `normalized_reference_path_segment` — converts extension slice segments like `extension:encounter` to `extension.where(url='http://...')` by looking up the URL from `must_supports[:extensions]` in the profile metadata.
  - `extension_url_for_reference_path` — looks up extension URL from profile metadata `must_supports[:extensions]`.
  - `extension_url_without_version` — strips version suffix from profile URLs before use in navigation paths.
  - `reference_navigation_context` — factory method for `ReferenceNavigationContext`.

- Removes manual `Claim.extension:encounter` FHIRPath workaround from `handle_claim_profile`. The navigation module now resolves this automatically:
```
  Claim.extension:encounter
    → normalized_reference_path_segment looks up must_supports[:extensions]
    → finds { id: "Claim.extension:encounter",
               url: "http://hl7.org/fhir/5.0/StructureDefinition/extension-Claim.encounter" }
    → produces: extension.where(url='http://hl7.org/fhir/5.0/...')
    → resolve_path navigates to Encounter/SurgicalEncounterExample
    → profile-encounter assigned 
```

- Updates `handle_claim_profile` requestedService path to use `value[x]` suffix for compatibility with the navigation module's choice type handling:
```
  # Before
  path: 'Claim.item.extension.value'

  # After
  path: 'Claim.item.extension:requestedService.value[x]'
```
  The navigation module normalizes this to:
```
  item.extension.where(url='...requestedService...').value[x]
```

### Operation Type Helpers

- Adds `submit_operation?`, `inquire_operation?`, and `request_operation` 
  helper methods to replace direct string comparisons of `request_type`.

- **Fixes a bug** where `request_type == 'submit'` comparisons were 
  silently failing because `request_type` is a compound string like 
  `submit_request` — not just `submit`. This caused incorrect validation 
  path selection and resources falling through to US Core profiles 
  incorrectly. The new helpers correctly extract the operation type:
```ruby
  def request_operation(request_type)
    request_type.to_s.split('_').first  # "submit_request" → "submit"
  end
```

---

### Before
- Bundle traversal depended on external FHIRPath evaluation.
- `Claim.extension:encounter` traversal required a manual FHIRPath workaround in `handle_claim_profile`.
- Extension slice traversal was not metadata-driven.
- `request_type == 'submit'` comparisons were silently failing for compound request type strings.

### After
- Bundle traversal uses shared Inferno core navigation, in-process.
- Extension slice traversal is resolved automatically through extracted metadata.
- `Encounter/SurgicalEncounterExample` is discovered through metadata-driven navigation without any manual workaround.
- Operation type detection correctly handles compound request type strings.

---

## Unit Testing

```sh
bundle exec rspec
bundle exec rubocop
```

---

## UI Testing Steps

Use the v2.2 client and server suites run against each other locally.

### What To Verify

**PAS profile precedence — must NOT fall back to US Core:**
- `Encounter/SurgicalEncounterExample` → validates against `profile-encounter`. No `6.1.0` reference should appear.
- `ServiceRequest/ReferralRequestExample` → validates against `profile-servicerequest`. No `us-core-servicerequest` should appear.
- `Patient/SubscriberExample` → validates against `profile-subscriber`.

**US Core via declared meta.profile:**
- `DocumentReference/episode-summary` → validates against `us-core-documentreference|6.1.0`.

**General:**
- No resource that has a PAS profile should fall through to US Core.
- No new errors should appear that were not present before this change.

### Regression Check

Run the **PAS v2.0.1 suite** and confirm:
- No `us-core-*` profiles appear in validation messages.
- No `6.1.0` references appear anywhere.
- Behavior is identical to before this change.

---

## Notes

This PR only changes traversal and navigation handling, and operation profile selection logic. The existing validation logic, profile precedence order, US Core fallback behavior, and bundle entry profile map were not modified.

### DIFF
--- a/lib/davinci_pas_test_kit/cross_suite/pas_bundle_validation.rb
+++ b/lib/davinci_pas_test_kit/cross_suite/pas_bundle_validation.rb
@@ -325,7 +325,9 @@ module DaVinciPASTestKit
           end
         end
 
-        validation_error_messages << generate_non_conformance_message(item) unless success_profile
+        next if success_profile
+
+        validation_error_messages << generate_non_conformance_message(item, base_profile, version)
       end
     end
 
@@ -479,9 +481,21 @@ module DaVinciPASTestKit
       end
     end
 
-    def generate_non_conformance_message(item)
+    def generate_non_conformance_message(item, base_profile, version)
+      target_profiles = item[:profile_urls].map { |url| display_target_profile_url(url, base_profile, version) }.uniq
       "#{item[:resource].resourceType}/#{item[:resource].id} is not conformant to any of the " \
-        "target profiles: #{item[:profile_urls]}."
+        "target profiles: #{target_profiles}."
+    end
+
+    # Renders a stored target profile URL for display in non-conformance messages. Mirrors how the profile
+    # was actually validated (see #profile_url_for_validation) so the listed profiles consistently carry a
+    # version, regardless of whether they were collected from versioned metadata, the unversioned
+    # requestedService reference list, or a declared meta.profile. The collected map is left untouched, so
+    # metadata lookups that rely on #profile_url_without_version are unaffected.
+    def display_target_profile_url(url, base_profile, version)
+      return base_profile if url == BASE_R4_PROFILE
+
+      profile_url_for_validation(url, base_profile, version)
     end
 
     # Processes each entry in a FHIR bundle to extract resource and possible profiles to validate against.
--- a/lib/davinci_pas_test_kit/cross_suite/pas_bundle_validation.rb
+++ b/lib/davinci_pas_test_kit/cross_suite/pas_bundle_validation.rb
@@ -1,5 +1,7 @@
 # frozen_string_literal: true
 
+require 'inferno/dsl/fhir_resource_navigation'
+
 require_relative '../parameters_helper'
 require_relative 'validation_test'
 require_relative 'pas_constants'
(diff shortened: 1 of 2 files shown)


karlnaden:
converting to draft - need to re-evaluate whether this is necessary or desirable.

PR #72, opened 2026-06-10
ID-148 Make non-ClaimResponse must support tests optional for v2.2.1 client …

FlexonyoPizza:
# Summary
Made non-ClaimResponse must support tests optional for v2.2.1 client suite in:
- pas_client_inquire_must_support_group.rb
- pas_client_inquire_response_must_support_group.rb
- pas_client_submit_must_support_group.rb
- pas_client_submit_response_must_support_group.rb

### DIFF
--- a/lib/davinci_pas_test_kit/client/generated/v2.2.1/pas_client_inquire_response_must_support_group.rb
+++ b/lib/davinci_pas_test_kit/client/generated/v2.2.1/pas_client_inquire_response_must_support_group.rb
@@ -39,9 +39,7 @@ module DaVinciPASTestKit
       test from: :pas_client_v221_inquire_response_must_support_pas_inquiry_response_bundle do
         optional
       end
-      test from: :pas_client_v221_inquire_response_must_support_claiminquiryresponse do
-        config(options: { user_input_validation: false })
-      end
+      test from: :pas_client_v221_inquire_response_must_support_claiminquiryresponse
       test from: :pas_client_v221_inquire_response_must_support_insurer do
         optional
       end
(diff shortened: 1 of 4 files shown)


lib/davinci_pas_test_kit/client/generated/v2.2.1/pas_client_inquire_response_must_support_group.rb
```diff
@@ -36,14 +36,30 @@ class PASClientInquireResponseMustSupportGroup < Inferno::TestGroup
...
-      test from: :pas_client_v221_inquire_response_must_support_requestor
-      test from: :pas_client_v221_inquire_response_must_support_beneficiary
-      test from: :pas_client_v221_inquire_response_must_support_practitioner
-      test from: :pas_client_v221_inquire_response_must_support_practitioner_role
-      test from: :pas_client_v221_inquire_response_must_support_task
+      test from: :pas_client_v221_inquire_response_must_support_pas_inquiry_response_bundle do
+        optional
+      end
+      test from: :pas_client_v221_inquire_response_must_support_claiminquiryresponse do
+        config(options: { user_input_validation: false })
```
ljtucker: Why did this particular line need a change? What does setting this config do differently than just allowing this test to run as before?

FlexonyoPizza: In must_support_test.rb, there's this if/else condition:
 
if user_input_validation
  skip_if resources_of_interest.blank?, "No #{resource_type} resources were found"
else
  assert resources_of_interest.present?, "No #{resource_type} resources were found"
end

which essentially allowed those claimResponse tests to skip without penalty if the tester didn't provide that bundle. If you change that config setting and set them manually to false for the claimresponse tests, they hit the else condition that asserts a failure if no resources are found.

ljtucker (changes requested):

ljtucker (approved):

karlnaden (changes requested):
Claim Update (10.3.03) and Claim Inquiry (10.6.02) must supports should be required

karlnaden (approved):
looks good now

PR #73, opened 2026-06-12
ID-152 Add metadata endpoint with a CapabilityStatement

FlexonyoPizza:
# Summary
The MockPASServer module in mock_pas_server.rb contains a FHIR CapabilityStatement from pas_server_capability_statement.json when a client requests the payer server's metadata endpoint. At boot time, the v2.2.1 client suite registers a route binding that URL to a lambda that fetches the capability_statement_response. When a request arrives, the method reads the JSON file from disk on first call (caching it in memory for subsequent requests) and returns a standard HTTP 200 response with a FHIR-typed JSON body.


lib/davinci_pas_test_kit/client/v2.2.1/client_suite.rb
```diff
@@ -126,6 +127,15 @@ class ClientSuite < Inferno::TestSuite
         SMARTAppLaunch::MockSMARTServer.smart_server_metadata(id)
       })
 
+      # FHIR capability statement for the simulated PAS payer server, declaring
+      # subscription support as required by the PAS IG.
+      route(:get, FHIR_METADATA_PATH, lambda { |env|
+        MockPASServer.capability_statement_response(env)
+      })
+      route(:get, SESSION_FHIR_METADATA_PATH, lambda { |env|
```
ljtucker, line 135: Why have two routes that produce the exact same document?

ljtucker (changes requested):

lib/davinci_pas_test_kit/client/v2.2.1/client_suite.rb
FlexonyoPizza: My original intent was to just implement the one route: "FHIR_METADATA_PATH" which uses a generic URL and a token in the request header that inferno looks up to figure out which session it belongs to. Copilot suggested that I also implement a path-based route as well that includes the session ID directly in the URL so theres no token lookup needed since some clients could be configured that way.

ljtucker: I'd say that if there are other examples of this in this suite that you can point to, then let's keep it. But if this would be the only URL the suite is hosting that you dont need a token lookup for, then we should not include it.

karlnaden: Both routes are needed. The PAS client tests support a session-specific fhir server url and these will need a metadata endpoint. There might be a fancy way to have a single route cover both flavors, but not sure it is worth it.

ljtucker: In that case, approved!

FlexonyoPizza: Yes it looks like there are other endpoints in this suite that use both URL variants. In **client_urls.rb**, you can see other endpoints follow the same format like:

SUBMIT_PATH = "#{FHIR_PATH}/Claim/$submit".freeze
SESSION_SUBMIT_PATH = SESSION_PATH_PLACEHOLDER + SUBMIT_PATH

INQUIRE_PATH = "#{FHIR_PATH}/Claim/$inquire".freeze
SESSION_INQUIRE_PATH = SESSION_PATH_PLACEHOLDER + INQUIRE_PATH

FHIR_SUBSCRIPTION_PATH = "#{FHIR_PATH}/Subscription".freeze
SESSION_FHIR_SUBSCRIPTION_PATH = SESSION_PATH_PLACEHOLDER + FHIR_SUBSCRIPTION_PATH

ljtucker (approved):

lib/davinci_pas_test_kit/client/metadata/pas_server_capability_statement.json
```diff
@@ -0,0 +1,138 @@
...
+          "display" : "United States of America"
+        }
+      ]
+    }
+  ],
+  "kind" : "requirements",
+  "fhirVersion" : "4.0.1",
+  "format" : [
+    "json",
+    "xml"
```
karlnaden: remove - inferno doesn't support XML

lib/davinci_pas_test_kit/client/metadata/pas_server_capability_statement.json
```diff
@@ -0,0 +1,138 @@
...
+      "coding" : [
+        {
+          "system" : "urn:iso:std:iso:3166",
+          "code" : "US",
+          "display" : "United States of America"
+        }
+      ]
+    }
+  ],
+  "kind" : "requirements",
```
karlnaden: Needs to be an `instance` kind, which will require some additional stuff around elements `software` and `implementation`

```suggestion
  "kind" : "instance",
```

lib/davinci_pas_test_kit/client/metadata/pas_server_capability_statement.json
```diff
@@ -0,0 +1,138 @@
+{
+  "resourceType" : "CapabilityStatement",
+  "id" : "IntermediaryCapabilities",
+  "language" : "en",
+  "text" : {
```
karlnaden: we're going to need to alter the contents, so I think removing the `langauge` and `text` is necessary to prevent it from getting out of sync

lib/davinci_pas_test_kit/client/metadata/pas_server_capability_statement.json
```diff
@@ -0,0 +1,138 @@
...
+  },
+  "extension" : [
```
karlnaden: these extensions can all be removed as they are relevant to the published version, not Inferno's version

lib/davinci_pas_test_kit/client/metadata/pas_server_capability_statement.json
```diff
@@ -0,0 +1,138 @@
...
+          {
+            "url" : "http://hl7.org/fhir/StructureDefinition/structuredefinition-conformance-derivedFrom",
+            "valueCanonical" : "http://hl7.org/fhir/us/davinci-pas/ImplementationGuide/hl7.fhir.us.davinci-pas"
+          }
+        ]
+      }
+    }
+  ],
+  "url" : "http://hl7.org/fhir/us/davinci-pas/CapabilityStatement/IntermediaryCapabilities",
+  "version" : "2.2.1",
```
karlnaden: I think we don't want a canonical URL and version, remove.

lib/davinci_pas_test_kit/client/metadata/pas_server_capability_statement.json
```diff
@@ -0,0 +1,138 @@
+{
+  "resourceType" : "CapabilityStatement",
+  "id" : "IntermediaryCapabilities",
```
karlnaden: remove the id

lib/davinci_pas_test_kit/client/metadata/pas_server_capability_statement.json
```diff
@@ -0,0 +1,138 @@
...
+            "valueCanonical" : "http://hl7.org/fhir/us/davinci-pas/ImplementationGuide/hl7.fhir.us.davinci-pas"
+          }
+        ]
+      }
+    }
+  ],
+  "url" : "http://hl7.org/fhir/us/davinci-pas/CapabilityStatement/IntermediaryCapabilities",
+  "version" : "2.2.1",
+  "name" : "PASIntermediary",
+  "title" : "Intermediary PAS Capabilities",
```
karlnaden: Update name and title to pertain to the particular PAS suite.

lib/davinci_pas_test_kit/client/metadata/pas_server_capability_statement.json
```diff
@@ -0,0 +1,138 @@
...
+  "status" : "active",
+  "experimental" : false,
+  "date" : "2023-08-08",
```
karlnaden: Update to a current date.

lib/davinci_pas_test_kit/client/metadata/pas_server_capability_statement.json
```diff
@@ -0,0 +1,138 @@
...
+  "publisher" : "HL7 International / Financial Management",
+  "contact" : [
```
karlnaden: remove publisher and contact

lib/davinci_pas_test_kit/client/metadata/pas_server_capability_statement.json
```diff
@@ -0,0 +1,138 @@
...
+      "name" : "HL7 International / Financial Management",
+      "telecom" : [
+        {
+          "system" : "url",
+          "value" : "http://www.hl7.org/Special/committees/fm"
+        }
+      ]
+    }
+  ],
+  "description" : "Capabilities required for an Intermediary participating in a PAS Exchange.",
```
karlnaden: update description to match name/title (details of the specific suite)

lib/davinci_pas_test_kit/client/metadata/pas_server_capability_statement.json
```diff
@@ -0,0 +1,138 @@
...
+  "jurisdiction" : [
```
karlnaden: remove jurisdiction

karlnaden (changes requested):
CapabilityStatement cleanup needed.  For the `software` and `implementation` elements needed for a `instance` kind, see an example [here](https://hl7.org/fhir/R4/capabilitystatement-example.json.html).

lib/davinci_pas_test_kit/client/metadata/pas_server_capability_statement.json
```diff
@@ -0,0 +1,54 @@
+{
+  "resourceType" : "CapabilityStatement",
+  "name" : "DaVinciPASClientSuiteV221SimulatedPayerServer",
+  "title" : "Da Vinci PAS Client Suite v2.2.1 Simulated Payer Server",
+  "status" : "active",
+  "experimental" : false,
+  "date" : "2026-06-15",
+  "description" : "CapabilityStatement for the Da Vinci PAS Client Suite v2.2.1 simulated payer/intermediary server. Declares support for the PAS $submit and $inquire operations on Claim and Subscription create/update/delete so that clients under test can discover subscription support.",
+  "kind" : "instance",
```
karlnaden, line 9: with kind = "instance", you need to add the `software` and `implementation` elements. Something like the below.

```suggestion
  "kind" : "instance",
  "software": {
    "name": "DaVinciPASClientSuiteV221SimulatedPayerServer"
  },
  "implementation": {
    "description": "Inferno Da Vinci PAS Client Suite v2.2.1 Simulated Payer Server"
  },
```

karlnaden (changes requested):
one more addition

karlnaden (approved):

PR #74, opened 2026-06-15
ID-56: Add v2.2.1 Claim Update scenario tests (Client and Server suites)

abhinandan2012:
## Summary

Test the PAS v2.2.x Claim Update `$submit` scenarios in both the Client and Server suites, with example payloads so the Server suite can drive the full sequence against the Client suite without a real external client.

### What changed

Client Suite - new "Claim Updates" group nested under "PAS Workflows" (`pas_client_v221_claim_updates_group`):

- Four wait tests, one per submission in the sequence. Each uses a unique request tag, has a dedicated response input, auto-continues when its request is received, and never triggers a Subscription notification even when the configured response is pended.
  - `pas_client_v221_claim_update_initial_submit_test` - initial PAS request
  - `pas_client_v221_claim_update_add_item_submit_test` - update that adds an item
  - `pas_client_v221_claim_update_modify_cancel_submit_test` - update that modifies one item and cancels another
  - `pas_client_v221_claim_update_cancel_all_submit_test` - update that cancels the entire request (certificationType code 3 in `Claim.extension`, no items required)
- Eight verification tests that reload the tagged requests and validate the submitted Claim Update Bundles.

Server Suite - new "Claim Updates" section (`pas_server_v221_claim_updates`), nested under "Demonstrate Workflow Support":

- Inferno acts as the client, driving the same four-step sequence against the server under test. Each step validates the request Bundle, submits it, and validates the server's response Bundle.
- Two response-verification tests confirm the server's handling of updates.
- The section is wired into both the generated `server_suite.rb` and the `server_suite.rb.erb` template, gated to v2.2+ via the `claim_updates_group?` generator helper.

Preset:

- Added four example Claim Update `$submit` payloads to `config/presets/pas_server_v221_run_against_pas_client.json.erb` (the "Run Against the PAS Client Suite" preset). They form a conformant initial-to-cancel-all chain so the Server suite can drive the entire sequence against the Client suite with no external client.

Shared/support changes:

- `cross_suite/tags.rb` - unique per-submission tags (`CLAIM_UPDATE_INITIAL_TAG`, `CLAIM_UPDATE_ADD_ITEM_TAG`, `CLAIM_UPDATE_MODIFY_CANCEL_TAG`, `CLAIM_UPDATE_CANCEL_ALL_TAG`).
- `client/endpoints/claim_endpoint.rb` - honors a per-test `claim_update_tag` for tagging and a `suppress_notifications` option so wait tests never notify.
- `server/claim_operation_logic.rb` and `server/server_response_bundle_validation_test.rb` - allow an explicit `request_tag` for tagging and reloading a single submission, falling back to the existing use-case tag.
- `cross_suite/pas_bundle_validation.rb` - for spec-67, the existing reference-presence check now skips `Claim.related` so the deliberately omitted grandparent Claim is not flagged. No new presence logic was written; spec-67 reuses `check_presence_of_referenced_resources`.

### Spec coverage

Client Suite verification tests:

| Test ID | Requirement |
| --- | --- |
| `pas_client_v221_claim_update_referenced_claim_test` | spec-65 - updated Claim referenced in `Claim.related.claim` and included in the Bundle |
| `pas_client_v221_claim_update_grandparent_excluded_test` | spec-66 - if the referenced Claim is itself an update, its referenced Claim is not included |
| `pas_client_v221_claim_update_referenced_resources_test` | spec-67 - all other referenced resources included in the Bundle (reuses existing presence check) |
| `pas_client_v221_claim_update_entries_preserved_test` | spec-68 - all prior item and supportingInfo entries retained with preserved sequences |
| `pas_client_v221_claim_update_cancelled_entries_test` | spec-69 - cancelled entries carry the `infoCancelled` modifier extension (valueBoolean true) |
| `pas_client_v221_claim_update_cancelled_items_certtype_test` | spec-70 - cancelled items carry a `certificationType` extension with code 3 (Cancel) |
| `pas_client_v221_claim_update_changed_entries_test` | spec-71 - added, modified, or cancelled entries carry an `infoChanged` extension |
| `pas_client_v221_claim_update_info_changed_code_test` | spec-72 - `infoChanged` valueCode is consistent with the value set |

Server Suite verification tests:

| Test ID | Requirement |
| --- | --- |
| `pas_server_v221_claim_update_item_sequence_echo_test` | spec-34 - ClaimResponse items echo submitted `item.sequence` values |
| `pas_server_v221_claim_update_current_results_test` | spec-46 - ClaimResponse includes current results for all submitted items, including changed or cancelled |

### Testing

- `bundle exec rspec` - full suite green (287 examples, 0 failures).
- `bundle exec rubocop` - no offenses on changed files.
- Manual UI verification via the loopback presets (steps below).

---

## UI Testing Steps

These steps follow the approach in [docs/Running-Suites-Against-Each-Other.md](https://github.com/inferno-framework/davinci-pas-test-kit/blob/support-v2.2.0/docs/Running-Suites-Against-Each-Other.md), adapted to v2.2.1 and the new Claim Updates sections.

### Setup

Start Inferno and open the UI at `http://localhost:4567`, then open two browser windows.

Client Suite window:

1. Select the "Da Vinci Prior Authorization Support (PAS)" test kit and choose "Da Vinci PAS Client Suite v2.2.1".
2. Set the client security type to "Other Authentication" (no-auth).
3. Click "Start Testing".
4. From the Preset dropdown, select "Run Against the PAS Server Suite".

Server Suite window:x`x 

1. Select the same test kit and choose "Da Vinci PAS Server Suite v2.2.1".
2. Click "Start Testing".
3. From the Preset dropdown, select "Run Against the PAS Client Suite". This populates the server endpoint (the Client suite's `demo_execution` FHIR endpoint) and the four Claim Update payload inputs:
   - "PAS $submit Payload: Initial Claim"
   - "PAS $submit Payload: Update Adding an Item"
   - "PAS $submit Payload: Update Modifying and Canceling Items"
   - "PAS $submit Payload: Update Canceling the Entire Request"

### Run order

1. Client Suite: click "Run All Tests". Confirm the authentication and connectivity prompts. When execution reaches "PAS Workflows", the new "Claim Updates" group will wait for the four submissions.
2. Server Suite: run "1 Subscription Setup", then run the "2 Demonstrate Workflow Support" groups in order (2.1 Approval, 2.2 Denial, 2.3 Pended), then run "2.4 Claim Updates". Running 2.4 sends the four preset Bundles to the Client suite, which auto-continues the Client suite wait tests.

### Where to see the Claim Updates results

Server Suite, group "2.4 Claim Updates" (`pas_server_v221_claim_updates`):

| UI item | Test ID | What it shows |
| --- | --- | --- |
| 2.4.1 Initial submission | `pas_server_v221_claim_update_request_validation_initial`, `..._submit_initial`, `..._response_validation_initial` | Request Bundle conformant (simulation), submit succeeds, response Bundle conformant |
| 2.4.2 Update: add an item | `..._request_validation_add_item`, `..._submit_add_item`, `..._response_validation_add_item` | Same three checks for the add-item update |
| 2.4.3 Update: modify an item and cancel an item | `..._request_validation_modify_cancel`, `..._submit_modify_cancel`, `..._response_validation_modify_cancel` | Same three checks for the modify-and-cancel update |
| 2.4.4 Update: cancel the entire request | `..._request_validation_cancel_all`, `..._submit_cancel_all`, `..._response_validation_cancel_all` | Same three checks for the cancel-all update |
| 2.4.5.01 | `pas_server_v221_claim_update_item_sequence_echo_test` | spec-34: response items echo submitted item sequences |
| 2.4.5.02 | `pas_server_v221_claim_update_current_results_test` | spec-46: response includes current results for all submitted items |

Expected: 2.4.1.01 through 2.4.4.01 (request validation) pass, 2.4.x.02 (submit) and 2.4.x.03 (response validation) pass, and 2.4.5.01 and 2.4.5.02 pass.

Client Suite, navigate to "PAS Workflows" > "Claim Updates" (`pas_client_v221_claim_updates_group`):

- Subgroup "Submit the claim update sequence" - the four wait tests auto-continue as the Server suite sends each Bundle:
  - `pas_client_v221_claim_update_initial_submit_test`
  - `pas_client_v221_claim_update_add_item_submit_test`
  - `pas_client_v221_claim_update_modify_cancel_submit_test`
  - `pas_client_v221_claim_update_cancel_all_submit_test`
- Subgroup "Verify the claim update interactions" - the eight verification tests (spec-65 through spec-72) run against the reloaded requests. See the spec coverage table above for the test IDs.

Expected: the four wait tests pass (auto-continue on receipt with no Subscription notification), and the eight verification tests pass.

To inspect details for any test, expand it and use the Messages tab (validation results), the Requests tab (the submitted Bundle and the response), and the About tab (the requirement description).


lib/davinci_pas_test_kit/client/v2.2.1/workflows/pas_client_claim_update_submit_tests.rb
```diff
@@ -0,0 +1,238 @@
...
+                            "Input '#{input_title(response_input)}' must be valid JSON"
+        elsif response_input.present?
+          add_message('info', %(
+            No response provided in input '#{input_title(response_input)}'. Inferno will generate an
+            approved response to the $submit request from the submitted Claim.
+          ))
+        end
+
+        wait_identifier = session_wait_identifier(client_id, session_url_path)
+        submit_endpoint = session_endpont_url(:submit, client_id, session_url_path)
```
ljtucker: This looks like a typo, where it should read session_endpoint_url(...)

ljtucker (changes requested):

lib/davinci_pas_test_kit/client/v2.2.1/workflows/pas_client_claim_update_submit_tests.rb
abhinandan2012: Fixed the typo and updated the method name along with its references.

ljtucker (approved):

lib/davinci_pas_test_kit/server/v2.2.1/claim_updates/server_claim_update_response_tests.rb
```diff
@@ -0,0 +1,150 @@
...
+      id :pas_server_v221_claim_update_item_sequence_echo_test
+      title 'Server echoes item.sequence values in Claim Update responses (spec-34)'
+      description %(
+        Verifies [spec-34](https://hl7.org/fhir/us/davinci-pas/2.2.1/en/conformancedetails.html#ci-c-spec-34):
+        each item returned on the PAS ClaimResponse **SHALL** echo the same `item.sequence` as that same item had
+        on the Claim. This test reloads each Claim Update submission sent to the server (by its tag) and confirms
+        that every `ClaimResponse.item.itemSequence` in the server's response corresponds to an `item.sequence`
+        that was present on the submitted Claim.
+      )
+      verifies_requirements 'hl7.fhir.us.davinci-pas_2.2.1@spec-34'
```
karlnaden: Style-wise, I don't think we should include references to the requirement ids within the titles of tests or generally repeat the requirement text verbatim within the description - that is what the `verifies_requirements` links provide, when setup correctly (see other comment to get them working). 

Recommendations here and in other places that you've done this:
- remove the requirement id from the title
- update the description to at most paraphrase the requirement, e.g., for this one "The PAS IG [requires](https://hl7.org/fhir/us/davinci-pas/2.2.1/en/conformancedetails.html#ci-c-spec-34) that ClaimResponses echo the `item.sequence` values from the submitted Claim."

I feel most strongly about the title and a bit less about the description, so if you want to skip that part I think that's ok.

lib/davinci_pas_test_kit/server/v2.2.1/claim_updates/server_claim_update_response_tests.rb
```diff
@@ -0,0 +1,150 @@
...
+      include ServerClaimUpdateResponseValidation
+
+      id :pas_server_v221_claim_update_current_results_test
+      title 'Server returns current results for all submitted items in Claim Update responses (spec-46)'
+      description %(
+        Verifies [spec-46](https://hl7.org/fhir/us/davinci-pas/2.2.1/en/conformancedetails.html#ci-c-spec-46):
+        the returned ClaimResponse **SHALL** include the current results for all submitted items, including any
+        items changed or canceled since the original authorization request. Because a conformant Claim Update
+        carries all items from the original and prior updates (spec-68), this test reloads each Claim Update
+        submission sent to the server (by its tag) and confirms the server's ClaimResponse includes an
```
karlnaden: Style: The tester doesn't know about the internals of Inferno including details like the use of tags to "load" requests to make them available for evaluation by tests. So keep descriptions higher-level, such as "this test checks each Claim Update made during this group and confirms...".

lib/davinci_pas_test_kit/client/v2.2.1/workflows/pas_client_claim_update_submit_tests.rb
```diff
@@ -0,0 +1,238 @@
...
+      title 'Client submits an update that adds an item'
+      description %(
+        Inferno waits for an updated prior authorization `$submit` request that adds a new item to
+        the previously submitted Claim. The update is expected to reference the original Claim in
+        `Claim.related.claim` (and include it in the Bundle), retain all previously submitted item
+        and supportingInfo entries with their `sequence` values, and mark the newly added item with
+        an `infoChanged` extension. Upon receipt, Inferno returns the configured response and
+        continues automatically.
+      )
+      verifies_requirements 'hl7.fhir.us.davinci-pas_2.2.1@spec-63'
```
karlnaden, line 141: To get these annotations to work, you'll need to update the suite definition to reference the `hl7.fhir.us.davinci-pas_2.2.1` requirement set. Looks like it is already there, but uses the wrong actor - should be "PAS Client" and "PAS Server" for the 2.2.1 versions (changed from 2.0.1).

lib/davinci_pas_test_kit/server/v2.2.1/claim_updates/server_claim_update_response_tests.rb
```diff
@@ -0,0 +1,150 @@
...
+
+        steps.each do |step|
+          unexpected = step[:response_sequences] - step[:submitted_sequences]
+          assert unexpected.empty?,
+                 "In the server's response to the #{step[:label]}, the ClaimResponse returned " \
+                 "item.itemSequence value(s) #{unexpected.join(', ')} that were not present on the submitted " \
+                 'Claim. Each returned item SHALL echo an item.sequence from the submitted Claim.'
+        end
+
+        pass 'Every ClaimResponse item echoes an item.sequence from the submitted Claim Update.'
```
karlnaden: Since this is just a re-iteration of the description of the test, we don't need it. The typical pattern for tests is that it will pass if no asserts fail, so omitting this will result in a pass state but without the result message.

Remove all such end-of-test `pass` directives.

karlnaden (changes requested):
In this first round of review, I'm requesting some changes on test documentation style to make them consistent with existing tests. I haven't done a pass through this test kit, but in general, my hope it to follow the same [style recommendations as I wrote up for CRD](https://github.com/inferno-framework/davinci-crd-test-kit/wiki/Technical-Overview#naming-and-style-guidelines), so ideally new tests will follow these. I didn't do a close review of those yet, but worth checking the new tests against these guidelines as you make these requested updates.

I haven't looked at behavior yet and expect to have more comments when I do.

lib/davinci_pas_test_kit/server/v2.2.1/claim_updates/server_claim_update_response_tests.rb
abhinandan2012: Removed the requirement IDs from all the test titles and reworded the descriptions to paraphrase the requirement with a link instead of repeating the normative text, leaving the verifies_requirements annotations to carry the specifics.

lib/davinci_pas_test_kit/server/v2.2.1/claim_updates/server_claim_update_response_tests.rb
abhinandan2012: Reworded these to stay high-level.

lib/davinci_pas_test_kit/server/v2.2.1/claim_updates/server_claim_update_response_tests.rb
abhinandan2012: Removed all the trailing pass directives so the tests pass implicitly when no assertion fails.

lib/davinci_pas_test_kit/client/v2.2.1/workflows/pas_client_claim_update_submit_tests.rb
abhinandan2012: Updated the suite definitions to use the PAS-prefixed actors for the `hl7.fhir.us.davinci-pas_2.2.1` set: PAS Client on the client suite and PAS Payer on the server suite (that's the server-side actor name in the 2.2.1 requirements source - PAS Server isn't used). The annotations now resolve, and I regenerated the coverage CSVs.

lib/davinci_pas_test_kit/client/v2.2.1/workflows/pas_client_claim_update_verification_tests.rb
```diff
@@ -0,0 +1,517 @@
...
+
+        bundle.entry.map(&:resource).find { |resource| resource.is_a?(FHIR::Claim) }
+      end
+
+      # Builds ordered prior/current Claim comparisons across the detail chain, skipping any
+      # link where a request was not received.
+      def claim_update_detail_comparisons
+        claims = claim_update_detail_chain.map do |step|
+          bundle = claim_update_bundle(step[:tag])
+          { label: step[:label], claim: bundle && first_claim(bundle) }
```
karlnaden: The short circuit syntax is confusing and the first_claim method handles a nil bundle.

```suggestion
          { label: step[:label], claim: first_claim(bundle) }
```

lib/davinci_pas_test_kit/client/v2.2.1/workflows/pas_client_claim_update_verification_tests.rb
karlnaden, line 1: This file is big enough that I think it will be better to have a single file for the helpers and then separate files for each individual test. That's the more common approach in Inferno and will keep the files small and contained.

lib/davinci_pas_test_kit/client/v2.2.1/workflows/pas_client_claim_update_submit_tests.rb
karlnaden, line 1: I think I'm ok having the abstract and 4 individual tests in a single file. There are other ways to do some of this with configuration and input renaming, but I don't think its worth re-writing it.

lib/davinci_pas_test_kit/client/v2.2.1/workflows/pas_client_claim_update_verification_tests.rb
```diff
@@ -0,0 +1,517 @@
...
+
+        steps.each do |step|
+          references = related_claim_references(step[:claim])
+          assert references.present?,
+                 "The #{step[:label]} Claim must reference the Claim being updated in Claim.related.claim."
+
+          references.each do |reference|
+            referenced_entry = bundle_entry_for_reference(step[:bundle], reference)
+            assert referenced_entry&.resource.is_a?(FHIR::Claim),
+                   "The Claim referenced in Claim.related.claim (#{reference}) of the #{step[:label]} " \
```
karlnaden: use the `add_message` and then `assert_no_error_messages` [pattern](https://inferno-framework.github.io/docs/writing-tests/assertions-and-results.html#adding-messages-directly-to-results) so that you don't stop evaluating at the first error, hiding others.

lib/davinci_pas_test_kit/client/v2.2.1/workflows/pas_client_claim_update_verification_tests.rb
```diff
@@ -0,0 +1,517 @@
...
+          next unless parent_claim.is_a?(FHIR::Claim)
+
+          grandparent_reference = related_claim_references(parent_claim).first
+          next if grandparent_reference.blank? # the referenced Claim is not itself an update
+
+          step.merge(parent_reference:, grandparent_reference:)
+        end
+
+        skip_if applicable.empty?,
+                'No Claim Update referencing another Claim Update was received, so this requirement does not apply.'
```
karlnaden:
```suggestion
                'No Claim Update referencing another Claim Update was received, so this requirement was not checked.'
```

lib/davinci_pas_test_kit/client/v2.2.1/workflows/pas_client_claim_update_verification_tests.rb
```diff
@@ -0,0 +1,517 @@
...
+
+        applicable.each do |step|
+          grandparent_entry = bundle_entry_for_reference(step[:bundle], step[:grandparent_reference])
+          assert grandparent_entry.nil?,
```
karlnaden: use the `add_message` and then `assert_no_error_messages` [pattern](https://inferno-framework.github.io/docs/writing-tests/assertions-and-results.html#adding-messages-directly-to-results) so that you don't stop evaluating at the first error, hiding others.

lib/davinci_pas_test_kit/client/v2.2.1/workflows/pas_client_claim_update_verification_tests.rb
```diff
@@ -0,0 +1,517 @@
...
+        claims.each_cons(2).filter_map do |prior, current|
+          next if prior[:claim].nil? || current[:claim].nil?
+
+          { prior: prior[:claim], current: current[:claim],
+            prior_label: prior[:label], current_label: current[:label] }
+        end
+      end
+
+      # Finds the bundle entry that a reference points to, accommodating absolute fullUrls,
+      # urn:uuid fullUrls, and relative "ResourceType/id" references.
```
karlnaden:
```suggestion
      # Finds the bundle entry that a reference points to, accommodating absolute fullUrls,
      # urn:uuid fullUrls, and relative "ResourceType/id" references.
      # NOTE: purposefully more permissive than strict FHIR Bundle reference resolution logic
      # to allow for evaluation. Other steps will catch issues like missing fullUrl values.
```

lib/davinci_pas_test_kit/client/v2.2.1/workflows/pas_client_claim_update_verification_tests.rb
```diff
@@ -0,0 +1,517 @@
...
+        steps.each do |step|
+          bundle = step[:bundle]
+          base_url = extract_base_url(bundle.entry.first&.fullUrl)
+          validation_error_messages.clear
+          # The presence of the Claim being updated (Claim.related.claim) is governed by spec-65/66,
+          # which also require that an updated Claim's own referenced Claim be omitted. Excluding
+          # `related` here scopes the reused reference-presence check to the "other" referenced
+          # resources and avoids recursing into the deliberately-omitted grandparent Claim.
+          check_presence_of_referenced_resources(claim_without_related(step[:claim]), base_url, bundle.entry)
+          assert validation_error_messages.blank?,
```
karlnaden: use the `add_message` and then `assert_no_error_messages` [pattern](https://inferno-framework.github.io/docs/writing-tests/assertions-and-results.html#adding-messages-directly-to-results) so that you don't stop evaluating at the first error, hiding others.

lib/davinci_pas_test_kit/client/v2.2.1/workflows/pas_client_claim_update_verification_tests.rb
```diff
@@ -0,0 +1,517 @@
...
+      verifies_requirements 'hl7.fhir.us.davinci-pas_2.2.1@spec-68'
+
+      run do
+        comparisons = claim_update_detail_comparisons
+        skip_if comparisons.empty?,
+                'Not enough consecutive Claim Update requests were received to verify entry preservation.'
+
+        comparisons.each do |comparison|
+          missing_items = entry_sequences(comparison[:prior].item) - entry_sequences(comparison[:current].item)
+          assert missing_items.empty?,
```
karlnaden: use the `add_message` and then `assert_no_error_messages` [pattern](https://inferno-framework.github.io/docs/writing-tests/assertions-and-results.html#adding-messages-directly-to-results) so that you don't stop evaluating at the first error, hiding others.

lib/davinci_pas_test_kit/client/v2.2.1/workflows/pas_client_claim_update_verification_tests.rb
```diff
@@ -0,0 +1,517 @@
...
+            .map { |entry| { step:, entry: } }
+        end
+
+        skip_if cancelled_entries.empty?,
+                'No canceled item or supportingInfo entries were received to verify (the modify-and-cancel ' \
+                'update is expected to cancel an item).'
+
+        cancelled_entries.each do |cancelled|
+          entry = cancelled[:entry]
+          assert info_cancelled?(entry),
```
karlnaden: use the `add_message` and then `assert_no_error_messages` [pattern](https://inferno-framework.github.io/docs/writing-tests/assertions-and-results.html#adding-messages-directly-to-results) so that you don't stop evaluating at the first error, hiding others.

lib/davinci_pas_test_kit/client/v2.2.1/workflows/pas_client_claim_update_verification_tests.rb
```diff
@@ -0,0 +1,517 @@
...
+            .map { |item| { step:, item: } }
+        end
+
+        skip_if cancelled_items.empty?,
+                'No canceled items were received to verify (the modify-and-cancel update is expected to ' \
+                'cancel an item).'
+
+        cancelled_items.each do |cancelled|
+          item = cancelled[:item]
+          assert certification_type_cancel?(item),
```
karlnaden: use the `add_message` and then `assert_no_error_messages` [pattern](https://inferno-framework.github.io/docs/writing-tests/assertions-and-results.html#adding-messages-directly-to-results) so that you don't stop evaluating at the first error, hiding others.

lib/davinci_pas_test_kit/client/v2.2.1/workflows/pas_client_claim_update_verification_tests.rb
```diff
@@ -0,0 +1,517 @@
...
+            if prior_entry.nil?
+              'added'
+            elsif newly_cancelled?(entry, prior_entry)
+              'canceled'
+            elsif entry_modified?(entry, prior_entry)
+              'modified'
+            end
+          next if change_type.nil?
+
+          assert info_changed_extension(entry).present?,
```
karlnaden: use the `add_message` and then `assert_no_error_messages` [pattern](https://inferno-framework.github.io/docs/writing-tests/assertions-and-results.html#adding-messages-directly-to-results) so that you don't stop evaluating at the first error, hiding others.

lib/davinci_pas_test_kit/client/v2.2.1/workflows/pas_client_claim_update_verification_tests.rb
```diff
@@ -0,0 +1,517 @@
...
+                'Not enough consecutive Claim Update requests were received to verify change marking.'
+
+        comparisons.each do |comparison|
+          assert_changed_entries_marked(comparison[:current].item, comparison[:prior].item, 'item', comparison)
+          assert_changed_entries_marked(comparison[:current].supportingInfo, comparison[:prior].supportingInfo,
+                                        'supportingInfo', comparison)
+        end
+      end
+
+      def assert_changed_entries_marked(current_entries, prior_entries, kind, comparison)
```
karlnaden:
```suggestion
          check_changed_entries_marked(comparison[:current].supportingInfo, comparison[:prior].supportingInfo,
                                        'supportingInfo', comparison)
        end
      end

      def check_changed_entries_marked(current_entries, prior_entries, kind, comparison)
```

lib/davinci_pas_test_kit/client/v2.2.1/workflows/pas_client_claim_update_verification_tests.rb
```diff
@@ -0,0 +1,517 @@
...
+          next if change_type.nil?
+
+          assert info_changed_extension(entry).present?,
+                 "The #{change_type} #{kind} (sequence #{entry.sequence}) in the #{comparison[:current_label]} " \
+                 "must contain an infoChanged extension (it changed relative to the #{comparison[:prior_label]})."
+        end
+      end
+    end
+
+    # spec-72
```
karlnaden: Combine this with the previous test. You can check multiple requirements in a single test and in particular here, I think it makes sense to combine these so that you're checking the presence and the value together.

lib/davinci_pas_test_kit/client/v2.2.1/client_suite.rb
karlnaden, line 1: This change didn't full take, I regenerated the requirements files and they are there now.

lib/davinci_pas_test_kit/cross_suite/pas_bundle_validation.rb
```diff
@@ -776,6 +776,7 @@ def check_presence_of_referenced_resources(target_resource, base_url, resources_
       else
         target_resource.source_hash.each_key do |attr|
           next if claim_response_request_attr?(target_resource, attr)
+          next if claim_related_attr?(target_resource, attr)
```
karlnaden: I don't think this is the right approach. By excluding it here, I think you skip the reference inclusion checks on the parent Claim instance, which I don't think we want to do - you might have something referenced there that isn't in the Update Claim.  That means you have to work harder to recognize a Claim that is not the primary one (I think you do it based on position in the Claim Update checks, maybe you can replicate it here) and exclude that one, but not all of them.

karlnaden (changes requested):
generally good - a few details to update. I still need to run a few more examples and observe some failures.

karlnaden (changes requested):
We still need a test for this bit even though it isn't listed as a formal requirement (created a HL7 JIRA ticket to make it one):

> To cancel the entire request, clients will send a [certificationType extension](https://hl7.org/fhir/us/davinci-pas/2.2.1/en/StructureDefinition-extension-certificationType.html) with a code of 3 (Cancel) in the Claim.extension element of a PAS Claim Update resource. No items are needed to be sent to cancel the entire authorization.

Add a verification test that checks the cancel request to make sure it meets this requirement.

lib/davinci_pas_test_kit/client/v2.2.1/workflows/pas_client_claim_update_verification_tests.rb
abhinandan2012: simplified

lib/davinci_pas_test_kit/client/v2.2.1/workflows/pas_client_claim_update_verification_tests.rb
abhinandan2012: split into a shared pas_client_claim_update_validation_utils.rb plus one file per test

lib/davinci_pas_test_kit/client/v2.2.1/workflows/pas_client_claim_update_submit_tests.rb
abhinandan2012: left as a single file

lib/davinci_pas_test_kit/client/v2.2.1/workflows/pas_client_claim_update_verification_tests.rb
abhinandan2012: changed

lib/davinci_pas_test_kit/client/v2.2.1/workflows/pas_client_claim_update_verification_tests.rb
abhinandan2012: updated

lib/davinci_pas_test_kit/client/v2.2.1/workflows/pas_client_claim_update_verification_tests.rb
abhinandan2012: changed

lib/davinci_pas_test_kit/client/v2.2.1/workflows/pas_client_claim_update_verification_tests.rb
abhinandan2012: updated

lib/davinci_pas_test_kit/client/v2.2.1/workflows/pas_client_claim_update_verification_tests.rb
abhinandan2012: changed

lib/davinci_pas_test_kit/client/v2.2.1/workflows/pas_client_claim_update_verification_tests.rb
abhinandan2012: changed

lib/davinci_pas_test_kit/client/v2.2.1/workflows/pas_client_claim_update_verification_tests.rb
abhinandan2012: changed

lib/davinci_pas_test_kit/client/v2.2.1/workflows/pas_client_claim_update_verification_tests.rb
abhinandan2012: changed

lib/davinci_pas_test_kit/client/v2.2.1/workflows/pas_client_claim_update_verification_tests.rb
abhinandan2012: updated

lib/davinci_pas_test_kit/client/v2.2.1/workflows/pas_client_claim_update_verification_tests.rb
abhinandan2012: combined into a single test

lib/davinci_pas_test_kit/client/v2.2.1/client_suite.rb
abhinandan2012: Thanks for regenerating those.

lib/davinci_pas_test_kit/cross_suite/pas_bundle_validation.rb
abhinandan2012: Reworked - replaced the blanket related skip with a skip_claim_related flag that's only set when recursing into a Claim reached via a reference (a non-primary Claim). The primary Update Claim's related is still traversed, so its parent and the parent's own referenced resources are checked; only the omitted grandparent is skipped.

lib/davinci_pas_test_kit/client/v2.2.1/workflows/pas_client_claim_update_verification_tests.rb
abhinandan2012: changed

lib/davinci_pas_test_kit/client/v2.2.1/workflows/pas_client_claim_update_verification_tests.rb
karlnaden: Descriptions generally good, but helpful to add a "See Messages for details." suffix (e.g., like [here](https://github.com/inferno-framework/davinci-crd-test-kit/blob/30ef77e79b3f9b2c5659007a9de728b509ad1550/lib/davinci_crd_test_kit/client/v2.2.1/auth/decode_auth_token_test.rb#L64) in CRD)

lib/davinci_pas_test_kit/cross_suite/pas_bundle_validation.rb
karlnaden: remove this line, you already know that `target_resource.reference.present?` is true from 767, so `ref.blank?` is always false.

lib/davinci_pas_test_kit/client/v2.2.1/workflows/pas_client_claim_update_verification_tests.rb
abhinandan2012: added

lib/davinci_pas_test_kit/cross_suite/pas_bundle_validation.rb
```diff
@@ -771,17 +777,23 @@ def check_presence_of_referenced_resources(target_resource, base_url, resources_
         end
 
         if matching_resources.length.positive?
-          check_presence_of_referenced_resources(matching_resources.first, base_url, resources_to_match)
+          # A resource reached by following a reference is an included resource, not the primary Claim
+          # being validated; if it is itself a Claim Update, its referenced grandparent Claim is omitted.
+          check_presence_of_referenced_resources(matching_resources.first, base_url, resources_to_match,
```
karlnaden: pre-existing bug that worked, but inefficiently - currently called on the Bundle.entry, should go right to the resource.

```suggestion
          check_presence_of_referenced_resources(matching_resources.first.resource, base_url, resources_to_match,
```

lib/davinci_pas_test_kit/client/v2.2.1/workflows/pas_client_claim_update_referenced_claim_test.rb
```diff
@@ -0,0 +1,46 @@
...
+
+      id :pas_client_v221_claim_update_referenced_claim_test
+      title 'Claim Update references and includes the Claim being updated'
+      description %(
+        The PAS IG [requires](https://hl7.org/fhir/us/davinci-pas/2.2.1/en/conformancedetails.html#ci-c-spec-65)
+        that the Claim being updated is referenced in the `Claim.related.claim` element and included in the
+        Bundle. This test checks each Claim Update made during this group and confirms that it references a
+        prior Claim and that the referenced Claim is present in the Bundle.
+      )
+      verifies_requirements 'hl7.fhir.us.davinci-pas_2.2.1@spec-65'
```
karlnaden: add [§prof-4](https://hl7.org/fhir/us/davinci-pas/2.2.1/en/StructureDefinition-profile-claim-update.html#ci-c-prof-4)

```suggestion
      verifies_requirements 'hl7.fhir.us.davinci-pas_2.2.1@spec-65', 'hl7.fhir.us.davinci-pas_2.2.1@prof-4'
```

lib/davinci_pas_test_kit/client/v2.2.1/workflows/pas_client_claim_update_entries_preserved_test.rb
```diff
@@ -0,0 +1,49 @@
...
+      id :pas_client_v221_claim_update_entries_preserved_test
+      title 'Claim Update preserves all prior item and supportingInfo entries'
+      description %(
+        The PAS IG [requires](https://hl7.org/fhir/us/davinci-pas/2.2.1/en/conformancedetails.html#ci-c-spec-68)
+        that when changing the details of the request, the Claim contains all item and supportingInfo entries
+        from the original Claim and any previous update Claims, with `sequence` values preserved. This test
+        compares each detail-bearing Claim Update against the immediately prior Claim and confirms that every
+        prior item and supportingInfo sequence is still present.
+      )
+      verifies_requirements 'hl7.fhir.us.davinci-pas_2.2.1@spec-68'
```
karlnaden: This and others have a corresponding "prof" requirement. Check that the following are added: 
- [§prof-5](https://hl7.org/fhir/us/davinci-pas/2.2.1/en/StructureDefinition-profile-claim-update.html#ci-c-prof-5)
- [§prof-6](https://hl7.org/fhir/us/davinci-pas/2.2.1/en/StructureDefinition-profile-claim-update.html#ci-c-prof-6)
- [§prof-7](https://hl7.org/fhir/us/davinci-pas/2.2.1/en/StructureDefinition-profile-claim-update.html#ci-c-prof-7)
- [§prof-8](https://hl7.org/fhir/us/davinci-pas/2.2.1/en/StructureDefinition-profile-claim-update.html#ci-c-prof-8)
- [§prof-9](https://hl7.org/fhir/us/davinci-pas/2.2.1/en/StructureDefinition-profile-claim-update.html#ci-c-prof-9)
- [§prof-10](https://hl7.org/fhir/us/davinci-pas/2.2.1/en/StructureDefinition-profile-claim-update.html#ci-c-prof-10)

lib/davinci_pas_test_kit/client/v2.2.1/pas_client_claim_updates_group.rb
karlnaden, line 1: I don't think that any of these tests verifies the conformance of the requests and responses (simulation). Add them - you should be able to use our standard PAS Bundle validation.

karlnaden (changes requested):
updates look good. A few more I found after another review.

lib/davinci_pas_test_kit/cross_suite/pas_bundle_validation.rb
abhinandan2012: removed dead return if ref.blank?

lib/davinci_pas_test_kit/cross_suite/pas_bundle_validation.rb
abhinandan2012: changed

lib/davinci_pas_test_kit/client/v2.2.1/workflows/pas_client_claim_update_referenced_claim_test.rb
abhinandan2012: added

lib/davinci_pas_test_kit/client/v2.2.1/workflows/pas_client_claim_update_entries_preserved_test.rb
abhinandan2012: added the prof requirements to the right tests.

lib/davinci_pas_test_kit/client/v2.2.1/pas_client_claim_updates_group.rb
abhinandan2012: added

karlnaden (approved):

PR #75, opened 2026-06-17
ID-149 Add must support attestation for v2.2.1 client request groups

FlexonyoPizza:
# Summary
## New Files
- **must_support_with_attestation_option.rb**
New iteration of the old must_support_attestation_test.rb test class that now lives solely in the v2.2.1 client suite folder. It takes a list of profiles and checks whether the client's requests included all the "must support" data elements for those profiles. Any element it didn't see, it writes out as an info note instead of appearing as a list of all the unobserved elements on the attestation message. If everything was seen, the test passes. If something was missing, it pauses and asks the tester to attest: "the client doesn't collect this data (and isn't required to)" - saying yes passes the test, saying no fails it. One class is reused for all the attestation tests; what differs between them is just the configured list of profiles.

- **must_support_with_attestation_option_spec.rb** that contains 5 new spec tests:
1. When every must support element is observed therefore the test passes immediately w/o attestation
2. When must support elements are not observed, it logs each unobserved element as an info message and waits for an attestation
3. When must support elements are not observed, passes when the tester attests true
4. When must support elements are not observed, fails when tester attests false
5. when require_one_of is true and none of the listed profiles are present

## Changed Files
- **must_support_data_gathering.rb (shared helper module)**
Migrated the x12 false positive correction logic from **must_support_test.rb** to here. The reason is because both the existing must-support test and the new attestation test need this exact correction, however neither is a subclass of the other, the one place they both share is this module, so the logic lives here and both call it. 

- **must_support_test.rb** 
Where the x12 correction logic that used to be written here was removed, and replaced it with a single call to the shared version above.

- **client_must_support_group_generator.rb  (writes the group files)**
Added a handful of small helper methods: a switch that turns the new behavior on only for v2.2.1, a check for which profiles must stay mandatory (Claim / ClaimResponse), a way to mark the other per-profile response tests as optional, and helpers that format a profile list into the configuration the new test expects.

- **client_must_support_group.rb.erb**
Split into two paths. For v2.2.1 it now writes the collapsed structure: one attestation test for "at least one request profile," the mandatory Claim test on its own, and one attestation test for "everything else." For v2.0.1 it writes the original layout

- **client_must_support_response_group.rb.erb**
A small addition that now marks each response test as optional for v2.2.1 (except ClaimResponse, which stays required)  to keep the response groups' optional structure intact when the files are regenerated, instead of losing it.

- **pas_client_submit_must_support_group.rb (10.3 tests)** --> now contains 3 tests:
**10.3.01:** the four request service profiles (DeviceRequest / MedicationRequest / NutritionOrder / ServiceRequest), bundled one attestation test with require_one_of: true (at least one must be present)
**10.3.02:** Claim Update --> mandatory
**10.3.03:** all the other supporting profiles (Coverage, Insurer, Requestor, Beneficiary, Subscriber, Practitioner, PractitionerRole, Bundle), bundled into one attestation test

- **pas_client_inquiry_must_support_group.rb (10.6 tests)** --> 2 tests:
**10.6.01:** Claim Inquiry mandatory
**10.6.02:** all the other supporting profiles, bundled into one attestation test

# Testing
- bundle exec rspec
- bundle exec rake pas:generate  - regenerates all the must-support group and per-profile files, run it to make sure the contents of your generated files stay preserved
- If you would like to see the attestation branch in action, you can manually remove must support elements in **must_support_with_attestation_option.rb** in the gather_unobserved_elements method like this:
### TEMP (testing only) - force elements to look unobserved; remove before commit
        (grouped_resources['Coverage'] || []).each do |coverage|
          coverage.period = nil
          coverage.subscriberId = nil
          coverage.type = nil
        end
        (grouped_resources['Practitioner'] || []).each do |practitioner|
          practitioner.telecom = nil
          practitioner.address = nil
        end
        (grouped_resources['Patient'] || []).each do |patient|
          patient.birthDate = nil
          patient.address = nil
        end
image


lib/davinci_pas_test_kit/client/generated/v2.2.1/pas_client_inquire_must_support_group.rb
```diff
@@ -33,31 +34,34 @@ class PASClientInquireMustSupportGroup < Inferno::TestGroup
         
       )
       run_as_group
-
-      test from: :pas_client_v221_inquire_request_must_support_pas_inquiry_request_bundle do
-        optional
-      end
+      
+      # Mandatory - the PAS Claim Inquiry profile must always be demonstrated.
```
ljtucker, line 30: This looks like a direct change to a generated file. For this to work, we need to update the generator file instead of the generated ones.

lib/davinci_pas_test_kit/client/generated/v2.2.1/pas_client_inquire_must_support_group.rb
```diff
@@ -33,31 +34,34 @@ class PASClientInquireMustSupportGroup < Inferno::TestGroup
         
       )
       run_as_group
-
-      test from: :pas_client_v221_inquire_request_must_support_pas_inquiry_request_bundle do
```
ljtucker: Looking at the request/sugestion in the ticket, we should still have the request bundle be its own test. One test for bundle, one test for claim, and one test for the rest. Then we would only need the attestation for "the rest"

lib/davinci_pas_test_kit/cross_suite/must_support/must_support_attestation_test.rb
```diff
@@ -0,0 +1,154 @@
...
+        missing = remove_x12_coverage_false_positives(missing, resources) if profile[:resource_type] == 'Coverage'
+
+        result[profile[:title] || profile[:resource_type]] = missing if missing.present?
+      end
+    end
+
+    # Coverage.relationship.coding:X12Code has an empty required binding discriminator values list,
+    # so Inferno cannot reliably detect it automatically. Approximate by checking for an x12.org
+    # system on Coverage.relationship.coding.
+    def remove_x12_coverage_false_positives(missing, resources)
```
ljtucker: Is this X12 handling code net new, or something spun up for this function? We do not want the behavior of the service changed in any way other than how we handle a failed check. The flow should be (for one single resource type):
1. The resource is checked exactly as it has been in current tests
2. If any elements inside the resource are missing, log warning messages that they are missing 
- This is a change from original behavior, as these had been getting logged as errors
3. At the end of the run, if at least one element was missing, show the attestation asking users to confrim whether or not they intended to leave those missing. Fail the test if not and pass it if yes  
- This is a brand new step

If you are grouping the tests together, then repeat steps 1 and 2 for all resource types in the group, keeping a list of elements missing for each, then have one attestation at the end. 

The attestation's list of missing elements should read like:
- Resource 1
  - Element 1
  - Element 2
  - Element 3
- Resource 2
  - Element 1

lib/davinci_pas_test_kit/cross_suite/must_support/must_support_attestation_test.rb
```diff
@@ -0,0 +1,154 @@
...
+        by the client:
+
+        #{element_list}
+
+        Attest that the client system does **not** collect the data for the unobserved elements
+        listed above (and is not required to under the PAS IG).
+
+        [Click here](#{attest_true_url}) if the above statement is **true**. The test will **pass**.
+
+        [Click here](#{attest_false_url}) if the above statement is **false**. The test will **fail**.
```
ljtucker: This is a brand new class/test and in turn needs its own spec tests to ensure it works correctly

ljtucker (changes requested):

lib/davinci_pas_test_kit/client/generated/v2.2.1/pas_client_inquire_must_support_group.rb
karlnaden: I think its ok to have Bundles under the rest. Bundle must support isn't all that interesting because all must support elements are required, so if they aren't demonstrated there will be other errors (e.g., see [PAS Request Bundle](https://hl7.org/fhir/us/davinci-pas/2.2.1/en/StructureDefinition-profile-pas-request-bundle.html)). The third list was request profiles, which is only present on submit.

lib/davinci_pas_test_kit/cross_suite/must_support/must_support_attestation_test.rb
```diff
@@ -0,0 +1,154 @@
...
+      if require_one_of
+        assert resources_of_interest.present?,
+               "#{type.titleize} Bundle(s) must include at least one instance of one of these " \
+               "resource types: #{target_resource_types.join(', ')}"
+      end
+
+      unobserved_by_profile = gather_unobserved_elements
+
+      unobserved_by_profile.each do |profile_title, missing|
+        add_message('info', %(
```
karlnaden: I think these will be easier to parse / use downstream if there's a single info message per element. Its a little speculative on my part, but here's how I think the messages should look:

"Unobserved must support element for profile <profile>: <element>"

I think I would not add any formatting like bolding. 

This means that:
- there is a list of unobserved elements in the test report
- an attestation of true indicates that all unobserved elements are not supported
- an attestation of false indicates that at least one unobserved element is supported. It doesn't tell you which, but I think that's fine. I envision this as eventual potential feedback on which must support elements are consistently not used (e.g., you could look across all Inferno PAS client 2.2.1 sessions over a given timeframe and generate a report)

lib/davinci_pas_test_kit/cross_suite/must_support/must_support_attestation_test.rb
karlnaden, line 1: When you have a shared bucket, there's always a tension around whether to preemptively put something there in case it might be used by the other suites. In this case, I don't think we have a planned or anticipated use case for using this kind of attestation option on server must support tests (that suite actually isn't going to be our responsibility anyway). So I think we should just put it in the client v2.2.1 tests and if and when this pattern gets used elsewhere, it can be factored out.

lib/davinci_pas_test_kit/client/v2.2.1/must_support/pas_client_must_support_attestation_test.rb
```diff
@@ -0,0 +1,13 @@
+require_relative '../../../cross_suite/must_support/must_support_attestation_test'
+require_relative '../urls'
+
+module DaVinciPASTestKit
+  module DaVinciPASV221
+    class PASClientMustSupportAttestationTest < DaVinciPASTestKit::MustSupportAttestationTest
+      include URLs
+
+      id :pas_client_v221_must_support_attestation
```
karlnaden: "must_support_attestation" sounds like it is a pure attestation per naming conventions elsewhere.  Let's instead use:
- ids: "must_support_with_attestation_option"
- class names: MustSupportWithAttestationOption
- titles: Keep them the same. Don't explicitly mention the attestation option.
- descriptions: can indicate here that testers will have the opportunity to attest that unobserved must support elements are not supported by their system.

lib/davinci_pas_test_kit/cross_suite/must_support/must_support_attestation_test.rb
karlnaden: I agree that the x12 stuff should leverage existing code rather than re-creating.

In terms of the rest of the comment, I think that is what the code is doing. @ljtucker is there a particular part of this that you think diverges from that description?

lib/davinci_pas_test_kit/client/generated/v2.2.1/pas_client_inquire_must_support_group.rb
```diff
@@ -33,31 +34,34 @@ class PASClientInquireMustSupportGroup < Inferno::TestGroup
...
-        optional
-      end
-      test from: :pas_client_v221_inquire_request_must_support_practitioner_role do
-        optional
+
+      # All other inquire request profiles - unobserved must support elements may be attested
+      # as not collected by the client system.
+      test from: :pas_client_v221_must_support_attestation do
+        id :pas_client_v221_inquire_request_other_must_support_attestation
+        title 'All other inquire request profile must support elements are observed (or attested)'
```
karlnaden:
```suggestion
        title 'All other inquire request profile must support elements are observed'
```

lib/davinci_pas_test_kit/client/generated/v2.2.1/pas_client_submit_must_support_group.rb
```diff
@@ -41,37 +42,55 @@ class PASClientSubmitMustSupportGroup < Inferno::TestGroup
...
-        optional
-      end
-      test from: :pas_client_v221_submit_request_must_support_pas_request_bundle do
-        optional
+      
+      # At least one of the PAS request profiles must be present; unobserved must support
+      # elements on the request profiles that are present may be attested as not collected.
+      test from: :pas_client_v221_must_support_attestation do
+        id :pas_client_v221_submit_request_profiles_must_support_attestation
+        title 'At least one request profile is observed with its must support elements (or attested)'
```
karlnaden:
```suggestion
        title 'At least one request profile is observed with its must support elements'
```

lib/davinci_pas_test_kit/client/generated/v2.2.1/pas_client_submit_must_support_group.rb
```diff
@@ -41,37 +42,55 @@ class PASClientSubmitMustSupportGroup < Inferno::TestGroup
...
-        optional
-      end
-      test from: :pas_client_v221_submit_request_must_support_practitioner_role do
-        optional
+
+      # All other submit request profiles - unobserved must support elements may be attested
+      # as not collected by the client system.
+      test from: :pas_client_v221_must_support_attestation do
+        id :pas_client_v221_submit_request_other_must_support_attestation
+        title 'All other submit request profile must support elements are observed (or attested)'
```
karlnaden:
```suggestion
        title 'All other submit request profile must support elements are observed'
```

lib/davinci_pas_test_kit/client/generated/v2.2.1/pas_client_inquire_must_support_group.rb
```diff
@@ -1,3 +1,4 @@
+require_relative '../../v2.2.1/must_support/pas_client_must_support_attestation_test'
 require_relative 'pas_inquiry_request_bundle/client_inquire_request_must_support_pas_inquiry_request_bundle_test'
```
karlnaden: This an other requires are no longer needed. In fact, many of these tests go away and may need to be deleted if they aren't used elsewhere.

lib/davinci_pas_test_kit/client/generated/v2.2.1/pas_client_submit_must_support_group.rb
```diff
@@ -1,4 +1,5 @@
 require_relative '../../v2.2.1/must_support/pas_client_must_support_request_profiles_test'
+require_relative '../../v2.2.1/must_support/pas_client_must_support_attestation_test'
 require_relative 'pas_request_bundle/client_submit_request_must_support_pas_request_bundle_test'
```
karlnaden: This an other requires are no longer needed. In fact, many of these tests go away and may need to be deleted if they aren't used elsewhere.

karlnaden (changes requested):
Broadly looks good, details need fine-tuning.

lib/davinci_pas_test_kit/client/v2.2.1/must_support/must_support_with_attestation_option.rb
```diff
@@ -0,0 +1,131 @@
...
+
+      id :pas_client_v221_must_support_with_attestation_option
+      title 'Must support elements are observed across requests'
+      description %(
+        This test reviews the must support elements observed across the listed profiles in the
+        requests made by the client. Each unobserved element is logged as an info message. If any
+        were not observed, the tester has the opportunity to attest that the client system does not
+        collect that data (and is not required to under the PAS implementation guide). When every
+        must support element was observed, the test passes without requiring any input.
+      )
```
karlnaden, line 26: This description has details that I think are best left out of the public-facing tests description:

```suggestion
      description %(
        This test reviews the must support elements observed across the listed profiles in the
        requests made by the client. If any
        were not observed, the tester has the opportunity to attest that the client system does not
        collect that data (and is not required to under the PAS implementation guide).
      )
```

Beyond that, I think it is helpful to have the profiles checked along with their must support elements listed, which would need to be done when adding the test. The previous version of these tests had that, and I don't think we want to regress if possible, though the descriptions. Looks like
- the description for the requests profiles are still in the (now unused and should be deleted) lib/davinci_pas_test_kit/client/v2.2.1/must_support/pas_client_must_support_request_profiles_test.rb
- descriptions for the "other" profiles are still in their generated must support tests, e.g. lib/davinci_pas_test_kit/client/generated/v2.2.1/pas_request_bundle/client_submit_request_must_support_pas_request_bundle_test.rb (also now unused an likely should be removed from the generator and the repo to avoid confusion).  Probably we copy / paste from those, but could also generate those lists somewhere and pull them into these tests, or maybe generate them at load time from the metadata. On the one hand, I don't want to expand the scope of this ticket, but I think its not a good idea to lose those element lists in the descriptions.

karlnaden (changes requested):

lib/davinci_pas_test_kit/client/v2.2.1/must_support/must_support_with_attestation_option.rb
karlnaden: Please also consider the rest of this comment around descriptions and additional code that is no longer used but is still present in the respository.

spec/davinci_pas_test_kit/client/v2.2.1/must_support_with_attestation_option_spec.rb
karlnaden, line 1: Spec cases to add:
1. require_one_of: false, no resources of the profile type submitted at all. This is the most important gap. It should fail or skip; currently it silently passes. A test would look like:
```
describe 'when require_one_of is false and no matching resources are in the session' do
  it 'does not silently pass' do
    # Bundle with no Coverage resource at all
    create_submit_request(FHIR::Bundle.new(type: 'collection').to_json)

    result = run(coverage_test)
    expect(result.result).not_to eq('pass')
  end
end
```

2. require_one_of: true with resources present but elements missing. The next if require_one_of && resources.blank? skip guard is completely untested. Need to verify that when DeviceRequest IS present with missing elements, it goes to attestation.

3. require_one_of: true with multiple profiles, only some present. The spec's require_one_of_test only lists one profile (DeviceRequest). The real submit group lists four. Testing that absent profiles are skipped while present ones are assessed is the core logic of that guard — and it's never exercised.

4. Multiple profiles in one test instance. Everything uses a single-profile configuration. The shared-resource-pool behavior (e.g., Insurer and Requestor both being resource_type: 'Organization') is untested.

5. No requests submitted at all (empty session). The current "hard fail" test sends a bundle with no DeviceRequest, but it does send a request. An empty session (load_tagged_requests returns nothing) is a distinct code path.

6. Attestation wait message content. The spec checks that a wait result has info messages, but never asserts on the wait message text itself — the markdown links, the profile name, the element names listed. A regression in attestation_message formatting would go undetected.

The most important to add is #1 (the bug) and #2 (the untested guard on the require_one_of path). The others add meaningful protection against regressions in code paths that currently have zero coverage.

lib/davinci_pas_test_kit/client/v2.2.1/must_support/must_support_with_attestation_option.rb
```diff
@@ -0,0 +1,137 @@
...
+
+        identifier = test_session_id
+        output attest_true_url: "#{resume_pass_url}?token=#{identifier}",
+               attest_false_url: "#{resume_fail_url}?token=#{identifier}"
+
+        wait(identifier:, message: attestation_message(unobserved_by_profile))
+      end
+
+      # @return unobserved must support elements keyed by profile title
+      def gather_unobserved_elements
```
karlnaden, line 110: This appears to not flag as problematic cases where no instances of a given resource are present. E.g., if I remove all PractitionerRole resources from the submit examples, the test passes without flagging PractitionerRole elements as missing.  Please fix that so that there is a failure. I think it would be ok to list the resource type as unobservered in the info messages and the attestation rather than listing all of the individual must support elements.

karlnaden (changes requested):

lib/davinci_pas_test_kit/generator/profile_metadata.rb
```diff
@@ -40,6 +40,33 @@ def delayed?
...
+          element_names << "#{element_paths} or #{extension_ids}"
+        end
+
+        (slice_names + element_names + extension_names)
+          .uniq
+          .sort
+          .map { |name| "#{' ' * indent}* #{name}" }
+          .join("\n")
+      end
+
```
FlexonyoPizza, line 69: This logic was originally in **must_support_test_generator.rb**, however I moved it here because its now needed by both the per-profile tests at generation time and the attestation test at load time. Same pattern with the x12 code before. MustSupportTestGenerator calls it with the method from here now.

lib/davinci_pas_test_kit/generator/templates/client_must_support_response_group.rb.erb
```diff
@@ -13,8 +13,14 @@ module DaVinciPASTestKit
...
-      test from: :<%= id %>
+      <%- profile_tests.each do |profile_test| -%>
+      <%- if profile_test[:optional] -%>
+      test from: :<%= profile_test[:id] %> do
+        optional
+      end
+      <%- else -%>
+      test from: :<%= profile_test[:id] %>
+      <%- end -%>
       <%- end -%>
```
FlexonyoPizza, line 24: This is the fix to one of the previous tasks where running **rake pas:generate** would strip the optional tags in the response groups

lib/davinci_pas_test_kit/client/v2.2.1/must_support/must_support_with_attestation_option.rb
karlnaden: Unused code gone and descriptions look better, but you still made a lot of changes to the name and description of the requests
- For the new 10.3.1, changes from the old PasClientMustSupportRequestProfilesTest should be minimal
    - The title should not change and should be “At least one instance of a request profile (PAS Medication Request, PAS Service Request, PAS Device Request, or PAS Nutrition Order) is observed with all of its must support elements
    - The description should stay largely the same, with only changes to indicate the new attestation option and why that is an option. I’d just tweak slightly the following paragraphs:
        - 2nd paragraph (after the profile list): “System are allowed to support only the request profiles that fit their use cases. However, they must support at least one of them (because `Claim.item.extension:requestedService` is a must support element) and for any request profiles they support, they must be able to populate all of the defined must support elements as long as that data is collected within and displayed by the system.”
        - 3rd paragraph: "This test ensures that the submitted request bundles include at least one instance of a profile listed above. Then for each profile observed, it checks for the presence of each must support element defined in that profile. For must support elements not observed on an included request profile, testers can attest that they are not supported by the client system."
- For the collapsed tests, now in 10.3.3 and 10.6.2, take a modified description from the originals:
    - "PAS client systems are required to be able to populate must support elements representing data collected and displayed to users on instances of all profiles included in requests. This test checks all identified instances of the profiles listed below within requests sent by the client to ensure that the must support elements are observed. For those not observed, testers can attest that they are not supported by the client system."

lib/davinci_pas_test_kit/client/v2.2.1/must_support/must_support_with_attestation_option.rb
karlnaden: Looks better. In the case where no requests were received, e.g., for $inquire, I think we should say that rather than indicating that no resources of a particular type were received. See lib/davinci_pas_test_kit/client/v2.2.1/workflows/pas_client_inquire_request_bundle_validation_test.rb line 54 for how this looks.  I think you should update lib/davinci_pas_test_kit/cross_suite/must_support/must_support_data_gathering.rb line 61 to be an assert with an error message like the previous one (depending on the operation) so that these are consistent. I think that will work, though you should double check.

karlnaden (changes requested):
A few additional tweaks to documentation and edge cases to maintain consistency.

lib/davinci_pas_test_kit/client/v2.2.1/must_support/must_support_with_attestation_option.rb
FlexonyoPizza: sounds good; I ended up not putting the assert in fetch_tagged_resources - Right now MustSupportTest evaluates skip_if resources_of_interest.blank? for the user-input-validation (response) tests. When I was doing some testing, when fetch_tagged_resources raised an assert there, those skips ended up failing. 

Also made some adjustments to the spec tests to reflect that assert message in both groups as well as restoring those PasClientMustSupportRequestProfilesTest title and descriptions
image
image

karlnaden: ok, you can test for the requests in the test class then. You can't use the LoadTaggedRequests because you'll duplicate the association, so you will have to use methods directly on Inferno::Repositories::Requests.

FlexonyoPizza: ok, my thinking then is to write a new method (must_support_requests) in **must_support_data_gathering.rb** that will be used to perform the original request association, but also use that same method in the attestation class which will read directly from Inferno::Repositories::Requests. 

So when fetch_tagged_resources runs as part of a MustSupportTest --> must_support_requests resolves to the mixin's default association 
When the same fetch_tagged_resources runs as part of the attestation test --> must_support_requests resolves to the override in the attestation class

FlexonyoPizza: - 10.3/10.6 (attestation tests) - they use the override (direct read, no duplicated association) so requests tab isnt clickable

- 10.4/10.7 (response must-support tests) - these are MustSupportTest instances using the default (load_tagged_requests) so the Requests tabs are clickable

image
image
image
image

karlnaden (approved):

PR #76, opened 2026-06-22
Migrate Metadata Updates out of PAS

ljtucker:
Several metadata extractor enhancements were made to a custom extension to the service in the Davinci PAS test kit. These enhancements were found to be useful for all kits, so they have been moved to Inferno Core for direct use through the base class. This PR removes those extensions from this repo as they have become redundant.

NOTE: Do not merge until the new inferno core gem is released


karlnaden (commented):
I ran `bundle exec rake pas:generate` while pointed to the inferno-core update branch expecting that the metadata would be generated the same way after the transition. They were, but there were differences in each of the 4 pas_client_*_must_support_group files - looks like `optional` flags are removed. Is that a change committed to the generator that wasn't committed to the generated files? I wouldn't expect your changes to cause that.

PR #79, opened 2026-07-07
ID-165: Criteria-based response selection for PAS client must support testing

abhinandan2012:
## Summary

The client suites previously served the tester-provided must support response bundles (`ms_submit_responses` / `ms_inquire_responses` inputs) sequentially, returning the nth bundle for the nth request. The request counting logic was also broken (`Inferno::Repositories::Requests#find_named_request` was called with a keyword signature that does not exist), so in practice the first bundle was always returned.

This PR replaces the sequential mechanism with criteria-based selection modeled on the DTR test kit's response selection utilities, plus request-based value substitution modeled on the CRD test kit's token replacement:

- Each response input now accepts either a single JSON FHIR Bundle or a JSON list of Bundles.
- For each incoming `$submit` / `$inquire` request, Inferno returns the first Bundle in the list whose selection criteria all match the request. A Bundle with no criteria always matches. If nothing matches, Inferno generates a default response.
- Selection criteria are expressed as extensions in a top-level `extension` array of the Bundle JSON:
  - `urn:inferno:pas:request-range` - `valueString` of comma-separated request numbers or ranges (for example `"1-2,4"`), matched against the 1-based count of successful requests to the same operation within the test, counting the current one.
  - `urn:inferno:pas:inclusion-criteria` - `valueExpression.expression` holding a FHIRPath expression evaluated against the incoming request Bundle; matches when it evaluates to a single truthy value.
  - When both extensions are present, both must match (AND). Inferno strips these extensions before returning the Bundle.
- `{{fhirpath}}` tokens anywhere in the selected Bundle (for example `{{Bundle.entry.first().resource.identifier.first().value}}`) are replaced with the result of evaluating the expression against the incoming request Bundle before the response is returned.
- Each selection decision is logged (`Selected tester-provided response bundle 2 of 3 for $submit request #2.`), so testers can trace why a bundle was chosen.

Implementation notes for reviewers:

- Candidate bundles are handled as parsed JSON hashes rather than FHIR model instances because `FHIR::Bundle` does not support top-level extensions and `FHIR.from_contents` silently drops them. After selection, stripping, and token replacement, the bundle is round-tripped through the FHIR model and flows into the existing `update_tester_provided_response` logic unchanged.
- The approval / denial / pended single-response inputs are untouched; only the must support list inputs use the new mechanism.
- Existing presets are unaffected: all preset `ms_*_responses` values are single-entry lists with no criteria, which the new mechanism serves for every request exactly as the old (buggy) behavior did.

New files: `cross_suite/fhirpath_utils.rb`, `cross_suite/response_selection_utils.rb`, plus unit and request-level specs. Modified: `client/user_input_response.rb` (`response_bundles`), `client/endpoints/claim_endpoint.rb` (selection, token replacement, logging), `client/abstract_gather_must_support_test.rb` (test and input documentation).

## How to test in the UI

Prerequisites: run the app locally with the background services up (validator and FHIRPath service are required; inclusion criteria and token replacement call the FHIRPath service).

### 1. Backward compatibility via the suite-vs-suite loopback

Follow `docs/Running-Suites-Against-Each-Other.md` (client suite with preset "Run Against the PAS Server Suite", server suite with preset "Run Against the PAS Client Suite") through the Must Support step:

1. In the client session (Da Vinci PAS Client Suite v2.2.1, Other Authentication), run group `10 Must Support Elements` and leave test `10.1.01` (id `pas_client_v221_gather_must_support`) waiting.
2. In the server session, run groups `3.1 $submit Element Support` and `3.2 $inquire Element Support`, which send the must support requests to the client session endpoints.
3. Return to the client session and click the confirmation link in the `10.1.01` wait dialog, then complete the attestation links.

Expected: identical behavior to before this change. The preset provides a single no-criteria bundle per operation, so every request receives that bundle. Tests `10.2.01` (id `pas_client_v221_request_bundle_validation_test`), `10.3.x` (`pas_client_v221_submit_request_must_support_*`), and `10.4.x` (`pas_client_v221_submit_response_must_support_*`) pass as they did previously.

### 2. Criteria-based selection and token replacement

1. Start a new client session (Da Vinci PAS Client Suite v2.2.1, Other Authentication) and run the `1 Client Registration` group with a session URL path (for example `demo_execution`).
2. Open group `10 Must Support Elements` and click RUN TESTS. In the input dialog, note the rewritten description of the `Must Support $submit Response Bundles` input: it documents the single-Bundle-or-list format, both criteria extensions, and the token syntax. Set the input to a three-entry list built from the Bundle in the "Must Support Response Examples (v2.2.1)" preset:
   - Bundle 1: set `entry[0].resource.id` to `ms-response-for-request-1` and add
     ```json
     "extension": [
       { "url": "urn:inferno:pas:request-range", "valueString": "1" }
     ]
     ```
   - Bundle 2: set `entry[0].resource.id` to `ms-response-criteria-match` and add
     ```json
     "extension": [
       { "url": "urn:inferno:pas:inclusion-criteria",
         "valueExpression": {
           "language": "text/fhirpath",
           "expression": "Bundle.entry.resource.ofType(Claim).identifier.where(value='DEMO-REQUEST-2').exists()"
         } }
     ]
     ```
   - Bundle 3: set `entry[0].resource.id` to `ms-response-default` with no extensions.
   - In all three, set `entry[0].resource.preAuthRef` to `{{Bundle.entry.first().resource.identifier.first().value}}`.
3. Submit the dialog; test `10.1.01` waits and displays the session `$submit` endpoint.
4. Send three `$submit` requests to that endpoint (any PAS request bundle works; the loopback preset's approval payload is convenient). For the second request only, change the Claim's identifier value to `DEMO-REQUEST-2`.
5. Click the confirmation link in the `10.1.01` wait dialog.

Expected results:

| Request | Selected bundle (ClaimResponse.id)   | Selected because              | preAuthRef value        |
|---------|--------------------------------------|-------------------------------|-------------------------|
| 1       | ms-response-for-request-1            | request-range "1"             | request 1 identifier    |
| 2       | ms-response-criteria-match           | FHIRPath matches identifier   | DEMO-REQUEST-2          |
| 3       | ms-response-default                  | no criteria (default)         | request 3 identifier    |

Where to see the results:

- Test `10.2.01 Submit Request Bundle is valid`: open the test, switch to the REQUESTS tab, and the three POSTs to `.../fhir/Claim/$submit` are listed. Open DETAILS on any of them and expand Response Body: the `id` of the first entry shows which bundle was selected, the top-level `extension` array is absent (criteria stripped), and `preAuthRef` contains the value pulled from that request. Note that incoming requests are listed under the conformance and must support tests that reload them by tag, not under `10.1.01` itself.
- The Inferno server log prints one line per request, for example:
  ```
  Selected tester-provided response bundle 2 of 3 for $submit request #2.
  ```
  and `No tester-provided response bundle matched ...` when the default response is generated.
- Groups `10.3` and `10.4` then evaluate must support coverage against the requests and the selected responses as before.

The same behavior applies to `$inquire` via the `Must Support $inquire Response Bundles` input and tests `10.5` through `10.7`, and to the v2.0.1 client suite's Must Support Elements group, which shares this implementation.

## Notes and known limitations

- Token replacement is string-level (mirroring CRD): a replaced value containing a double quote or backslash could break the JSON, in which case the unparsed replacement string is returned as-is.
- Request-range counting only counts requests that received HTTP 200, which is a behavior refinement over the old count-everything logic.
- Invalid range strings (for example `"3-2"`) raise `Inferno::Exceptions::TestSuiteImplementationException`; the wording blames the test suite although the range comes from tester input. This mirrors the DTR utility.
- List entries must be JSON objects; the old code also tolerated pre-serialized JSON strings, but criteria extensions cannot be expressed on those.
- FHIRPath function calls in criteria and tokens require parentheses (`first()`, `exists()`); a bare `.first` evaluates to empty.

## Verification screenshots

Full run of the Must Support Elements group (v2.2.1 client suite, driven via the suite-vs-suite loopback described in `docs/Running-Suites-Against-Each-Other.md`) with the new selection mechanism in place. All groups pass, including 10.2.02, which validates the responses Inferno served from the tester-provided input:

Must Support Elements group passing

Submit Bundle Conformance detail, 10.2.01 and 10.2.02 passing

The $submit requests received during the must support test, listed under 10.2.01:

Requests list under 10.2.01

Request detail for one of the exchanges: the response body returned by Inferno is the tester-provided must support response bundle selected by the new mechanism:

Request detail with served response bundle

Example server log output when a multi-bundle input with selection criteria is provided (from the scenario in the testing steps above):

```
Selected tester-provided response bundle 1 of 3 for $submit request #1.
Selected tester-provided response bundle 2 of 3 for $submit request #2.
Selected tester-provided response bundle 3 of 3 for $submit request #3.
```


karlnaden (changes requested):
I thought I had confirmed that Bundles had a top-level extension property, but turns out [I was wrong](https://hl7.org/fhir/R4/bundle.html#notes). Options:
1. Continue to use the top-level extension property even though it is not valid FHIR.
2. Move to an extension at a lower level of the Bundle, like on `link` or a primitive extension on one of the other top-level elements.
3. Add the filters at a higher-level, allowing testers to wrap the Bundle in another json object if they want to add criteria.

None of these option feel great, but 3 seems the least bad.  Can you take a shot at that.

karlnaden (approved):
Some additional fixes to error handling will be included in a future PR, but for now this behavior works and is an improvement over the current code.

PR #83, opened 2026-07-14
ID-88: Narrow the x12 validator message filter and apply it to both Client and Server suites

abhinandan2012:
## Summary

The PAS IG binds many elements to proprietary X12 value sets whose definitions are not available to the terminology server. The server suites carried a broad filter that suppressed any validator message containing "x12" at any severity, while the client suites had no x12 filtering at all, so unresolvable x12 value set warnings surfaced in client suite results.

Investigation against the current validator (wrapper 1.0.79, validator core 6.9.11), using the IG's own example bundles, showed that x12 messages now come in exactly two shapes, both at warning severity with no errors remaining:

- `ValueSet 'https://valueset.x12.org/...' not found`
- `A definition for CodeSystem 'https://codesystem.x12.org/...' could not be found, so the code cannot be validated`

The exact-string `ValueSet <url> not found` entries already present in `SUPPRESSED_MESSAGES` (copied from the IG's published suppression list) stopped matching because newer validator versions quote the URL, which is why the warnings became visible in the client suites. The CodeSystem shape is still caught by the existing generic "could not be found, so the code cannot be validated" entry.

### What changed

- `lib/davinci_pas_test_kit/cross_suite/validator_suppressions.rb` - added a narrow regex, `%r{ValueSet 'https://valueset\.x12\.org/[^']+' not found}`, to the shared `SUPPRESSED_MESSAGES` constant. Since all four suites apply this constant in their `exclude_message` blocks, the suppression now applies consistently to client and server suites at every severity (the exclusion logic matches on message text only).
- `lib/davinci_pas_test_kit/server/generated/v2.0.1/server_suite.rb`, `lib/davinci_pas_test_kit/server/generated/v2.2.1/server_suite.rb`, and `lib/davinci_pas_test_kit/generator/templates/server_suite.rb.erb` - removed the broad `message.message.downcase.include?('x12')` clause. The server exclusion logic is now identical to the client suites.

Net behavior: resolution failures for X12 value sets and code systems remain suppressed everywhere, while actionable messages that merely mention an x12 URL (for example, a code-not-in-value-set finding) are no longer swallowed. Replaying recorded validator responses for the IG example bundles through the old and new logic produced identical classification for every real message, so there is no user-visible regression from the narrowing.

The suite and test descriptions already note that elements bound to X12 value sets are not validated, so no description changes were needed.

### Spec coverage

New spec `spec/davinci_pas_test_kit/cross_suite/validator_suppressions_spec.rb` exercises the actual `exclude_message` proc registered by each of the four suites (16 examples):

| Case | Expected |
| --- | --- |
| x12 `ValueSet '<url>' not found` (real message text) at error, warning, and info severity | excluded |
| x12 `A definition for CodeSystem '<url>' could not be found...` | excluded |
| Non-x12 `ValueSet '<url>' not found` (US Core URL) | not excluded |
| Code membership failure referencing an x12 value set URL | not excluded |

### Testing

- `bundle exec rspec` - 391 examples, 0 failures.
- `bundle exec rubocop` - no offenses.
- Manual UI verification via the loopback presets (steps below).

---

## UI Testing Steps

These steps follow [docs/Running-Suites-Against-Each-Other.md](https://github.com/inferno-framework/davinci-pas-test-kit/blob/support-v2.2.0/docs/Running-Suites-Against-Each-Other.md) ("Basic Execution (v2.2.1)").

### Setup

Start Inferno (`./run.sh`), open the UI at `http://localhost:4567`, and open two browser windows.

Client Suite window:

1. Choose "Da Vinci PAS Client Suite v2.2.1" with the "Other Authentication" client security type and click "Start Testing".
2. From the Preset dropdown, select "Run Against the PAS Server Suite".

Server Suite window:

1. Choose "Da Vinci PAS Server Suite v2.2.1" and click "Start Testing".
2. From the Preset dropdown, select "Run Against the PAS Client Suite".

### Execution

Run the paired groups in order per the doc, submitting the pre-filled preset inputs and responding to the "User Action Required" dialogs as described there:

1. Client group 1 (Client Registration), then client group 10 plus server group 1 (Subscription Setup).
2. Client 11.1 with server 2.1 (Approval Workflow).
3. Client 11.2 with server 2.2 (Denial Workflow).
4. Client 11.3 with server 2.3 (Pended Workflow).
5. Client 11.4 with server 2.4 (Claim Updates).
6. Client 12 with server 3 (Must Support Elements).
7. Client 13.1 with server 4 (Operation Failure).

### Where to verify the results

The change is visible in the Messages tab of the Bundle validation tests:

- Client suite: expand any "... Bundle is valid" test under groups 11.1 through 11.4 (for example, 11.4.2.01 "Initial submission: submit request Bundle is valid") and open the MESSAGES tab. Before this change, this tab listed multiple warnings of the form `ValueSet 'https://valueset.x12.org/...' not found`. After this change, those warnings are gone.
- Server suite: open the request and response Bundle validation tests inside groups 2.1 through 2.4 and group 3 and check the MESSAGES tab the same way. These were already free of x12 messages under the old broad filter and must remain so.
- Confirm the filter is not over-broad: unrelated messages (for example, "Example URLs are not allowed in this context" and other non-x12 warnings produced by the preset payloads) must still appear in the same tabs.

### Test IDs

Automated:

- `spec/davinci_pas_test_kit/cross_suite/validator_suppressions_spec.rb` (new, 16 examples)
- Full suite: `bundle exec rspec`

### DIFF
--- a/lib/davinci_pas_test_kit/cross_suite/validator_suppressions.rb
+++ b/lib/davinci_pas_test_kit/cross_suite/validator_suppressions.rb
@@ -18,6 +18,11 @@ module DaVinciPASTestKit
     # The NUBC revenue code system is proprietary as well and cannot be resolved. The test kit's
     # example data references it under both of the URLs the IG has used for it.
     %r{A definition for CodeSystem 'https://www\.nubc\.org/(revenue-code|CodeSystem/RevenueCodes)' could not be found, so the code cannot be validated},
+    # Because that code system is unavailable, membership in the PAS-defined value set built on it
+    # cannot be checked either. Claim.item.revenue is must-support (not required), so the IG examples
+    # leave it empty and the IG QA never hits this, but the must support tests populate it, so the
+    # code cannot be checked against the value set and the error is unactionable.
+    "None of the codings provided are in the value set 'AHA NUBC Revenue Value Set' (http://hl7.org/fhir/us/davinci-pas/ValueSet/AHANUBCRevenueCodes",
     # The IG references this draft code system with bindings stronger than example.
     'Reference to draft CodeSystem http://hl7.org/fhir/us/davinci-pas/CodeSystem/PASTempCodes',
     # IG defect: these extensions are marked must-support at a location that their own context
@@ -36,8 +41,10 @@ module DaVinciPASTestKit
     # must-support on ClaimResponse.item.
     %r{The extension http://hl7\.org/fhir/us/davinci-pas/StructureDefinition/extension-communicatedDiagnosis( v[\d.]+)? is not allowed to be used at this point \(this element is \[(?:BackboneElement, )?ClaimResponse\.item\]},
     # extension-productOrServiceCodeEnd: context lists Claim.item/ClaimResponse, but
-    # profile-servicerequest marks it must-support on ServiceRequest.
+    # profile-servicerequest marks it must-support on ServiceRequest and profile-claimresponse marks
+    # it must-support on ClaimResponse.addItem.
     %r{The extension http://hl7\.org/fhir/us/davinci-pas/StructureDefinition/extension-productOrServiceCodeEnd( v[\d.]+)? is not allowed to be used at this point \(this element is \[ServiceRequest\]},
+    %r{The extension http://hl7\.org/fhir/us/davinci-pas/StructureDefinition/extension-productOrServiceCodeEnd( v[\d.]+)? is not allowed to be used at this point \(this element is \[(?:BackboneElement, )?ClaimResponse\.addItem\]},
     # extension-itemAuthorizedProvider: context allows ExplanationOfBenefit/ClaimResponse.item only,
     # but profile-claimresponse marks it must-support on the ClaimResponse root.
     %r{The extension http://hl7\.org/fhir/us/davinci-pas/StructureDefinition/extension-itemAuthorizedProvider( v[\d.]+)? is not allowed to be used at this point \(this element is \[ClaimResponse\]},
--- a/lib/davinci_pas_test_kit/cross_suite/validator_suppressions.rb
+++ b/lib/davinci_pas_test_kit/cross_suite/validator_suppressions.rb
@@ -12,9 +12,11 @@ module DaVinciPASTestKit
   # rubocop:disable Layout/LineLength
   V221_SUPPRESSIONS = [
     # X12 value set and code system definitions are proprietary and not available to the validator,
-    # so resolution failures for them are not actionable by users.
-    %r{ValueSet 'https://valueset\.x12\.org/[^']+' not found},
-    %r{A definition for CodeSystem 'https://codesystem\.x12\.org/[^']+' could not be found, so the code cannot be validated},
+    # so resolution failures for them are not actionable by users. The IG references X12 URLs under
+    # both x12.org subdomains, and uses valueset.x12.org URLs as code system URLs in some places, so
+    # both patterns accept either subdomain.
+    %r{ValueSet 'https://(?:valueset|codesystem)\.x12\.org/[^']+' not found},
+    %r{A definition for CodeSystem 'https://(?:codesystem|valueset)\.x12\.org/[^']+' could not be found, so the code cannot be validated},
     # The NUBC revenue code system is proprietary as well and cannot be resolved. The test kit's
     # example data references it under both of the URLs the IG has used for it.
     %r{A definition for CodeSystem 'https://www\.nubc\.org/(revenue-code|CodeSystem/RevenueCodes)' could not be found, so the code cannot be validated},
--- a/lib/davinci_pas_test_kit/cross_suite/validator_suppressions.rb
+++ b/lib/davinci_pas_test_kit/cross_suite/validator_suppressions.rb
@@ -25,6 +25,10 @@ module DaVinciPASTestKit
     # leave it empty and the IG QA never hits this, but the must support tests populate it, so the
     # code cannot be checked against the value set and the error is unactionable.
     "None of the codings provided are in the value set 'AHA NUBC Revenue Value Set' (http://hl7.org/fhir/us/davinci-pas/ValueSet/AHANUBCRevenueCodes",
+    # These PAS-defined value sets draw all their members from X12 code systems, so membership
+    # cannot be checked and the failure is unactionable. Anchored on each value set's canonical URL.
+    %r{None of the codings provided are in the value set .*http://hl7\.org/fhir/us/davinci-pas/ValueSet/pas-pwk01-attachment-report-type-code},
+    %r{None of the codings provided are in the value set .*http://hl7\.org/fhir/us/davinci-pas/ValueSet/PASCommunicationRequestMedium},
     # The IG references this draft code system with bindings stronger than example.
     'Reference to draft CodeSystem http://hl7.org/fhir/us/davinci-pas/CodeSystem/PASTempCodes',
     # IG defect: these extensions are marked must-support at a location that their own context
--- a/lib/davinci_pas_test_kit/cross_suite/validator_suppressions.rb
+++ b/lib/davinci_pas_test_kit/cross_suite/validator_suppressions.rb
@@ -122,6 +122,10 @@ module DaVinciPASTestKit
       'X12278DiagnosisCodes', # ValueSet includes full ICD-10-CM plus unavailable code systems; validator cannot fully expand
       'X12278RequestedServiceType', # X12 code system included in this PAS-defined value set
       'X12278LocationType', # X12 code system included in this PAS-defined value set
+      # X12 value set definitions are proprietary and unavailable to the validator, so resolution
+      # failures for them are not actionable. Current validator versions quote the URL, which the
+      # exact-string entries from the IG suppression list above no longer match.
+      %r{ValueSet 'https://valueset\.x12\.org/[^']+' not found},
       'ValueSet https://valueset.x12.org/x217/005010/response/2010B/NM1/1/01/00/98 not found',
       "Error from http://tx.fhir.org/r4: Error:A definition for the value Set 'http://hl7.org/fhir/us/core/ValueSet/us-core-usps-state|3.1.1' could not be found",
       "A definition for CodeSystem 'https://www.cms.gov/Medicare/Coding/place-of-service-codes/Place_of_Service_Code_Set' could not be found, so the code cannot be validated",
--- a/lib/davinci_pas_test_kit/cross_suite/validator_suppressions.rb
+++ b/lib/davinci_pas_test_kit/cross_suite/validator_suppressions.rb
@@ -15,40 +15,37 @@ module DaVinciPASTestKit
     # so resolution failures for them are not actionable by users.
     %r{ValueSet 'https://valueset\.x12\.org/[^']+' not found},
     %r{A definition for CodeSystem 'https://codesystem\.x12\.org/[^']+' could not be found, so the code cannot be validated},
-    # The NUBC revenue code system is proprietary as well, so it can be neither resolved nor
-    # expanded within the PAS-defined value set built on it. Code membership failures against that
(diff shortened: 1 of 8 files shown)


ljtucker (approved):

karlnaden (changes requested):
I'm a bit uncomfortable with the level of message suppression that we're doing and particularly the pretty broad suppressions. The changes are directionally right, but I think we want to be a bit more aggressive in how we change the filtering, at least for the v2.2.1 suites.  Here's my recommendation:

1. for 2.2.1 suites: create a new filter list that is targeted at the two classes of x12 filters that you identified. You might also review some of the "extra" filters added after line 120 to see if they are still showing up in the 2.2.1 suites and include more targeted versions of them them if they are. Otherwise, don't include them (start minimal)
2. for the 2.0.1 suites: add the new 2.2.1 lists to the old list.

This will be a little bit more work, but I think will be worth it because we'll be sure that we're not suppressing valid messages.

lib/davinci_pas_test_kit/cross_suite/validator_suppressions.rb
```diff
@@ -1,5 +1,57 @@
...
+    "None of the codings provided are in the value set 'X12 278 Requested Service Type' (http://hl7.org/fhir/us/davinci-pas/ValueSet/X12278RequestedServiceType",
+    # The IG references this draft code system with bindings stronger than example.
+    'Reference to draft CodeSystem http://hl7.org/fhir/us/davinci-pas/CodeSystem/PASTempCodes',
+    # No real content
+    /Details for (.+) matching against profile/,
+    # IG defect: extension context definitions for these extensions don't match where the profiles
+    # place them as must support elements.
+    # extension-authorizationNumber and extension-administrationReferenceNumber context allows
+    # Claim.item only, but profile-claim-base marks them must-support on Claim.extension
+    %r{The extension http://hl7\.org/fhir/us/davinci-pas/StructureDefinition/extension-authorizationNumber( v[\d.]+)? is not allowed to be used at this point},
```
karlnaden: Looking at these suppressions around extension placement, it looks like we're not taking into account where it was seen. I think that's a mistake - in this case, we want to allow it to be used at the Claim level, but not additional places. Can you adjust this and other filters like it to target the specific locations that they need to be allowed but aren't currently.

lib/davinci_pas_test_kit/cross_suite/validator_suppressions.rb
```diff
@@ -1,5 +1,57 @@
...
+    # value set are false positives -- the IG's own examples fail the check.
+    "A definition for CodeSystem 'https://www.nubc.org/revenue-code' could not be found, so the code cannot be validated",
+    "None of the codings provided are in the value set 'AHA NUBC Revenue Value Set' (http://hl7.org/fhir/us/davinci-pas/ValueSet/AHANUBCRevenueCodes",
+    # Same situation for the PAS-defined X12278RequestedServiceType value set, which draws its
+    # members from unavailable X12 code systems.
+    "None of the codings provided are in the value set 'X12 278 Requested Service Type' (http://hl7.org/fhir/us/davinci-pas/ValueSet/X12278RequestedServiceType",
+    # The IG references this draft code system with bindings stronger than example.
+    'Reference to draft CodeSystem http://hl7.org/fhir/us/davinci-pas/CodeSystem/PASTempCodes',
+    # No real content
+    /Details for (.+) matching against profile/,
```
karlnaden: Is this showing up if it isn't filtered?

lib/davinci_pas_test_kit/cross_suite/validator_suppressions.rb
```diff
@@ -1,5 +1,57 @@
...
+  #
+  # rubocop:disable Layout/LineLength
+  V221_SUPPRESSIONS = [
+    # X12 value set and code system definitions are proprietary and not available to the validator,
+    # so resolution failures for them are not actionable by users.
+    %r{ValueSet 'https://valueset\.x12\.org/[^']+' not found},
+    %r{A definition for CodeSystem 'https://codesystem\.x12\.org/[^']+' could not be found, so the code cannot be validated},
+    # The NUBC revenue code system is proprietary as well, so it can be neither resolved nor
+    # expanded within the PAS-defined value set built on it. Code membership failures against that
+    # value set are false positives -- the IG's own examples fail the check.
```
karlnaden: > the IG's own examples fail the check.

I'm not seeing this error in the QA reports for either the [2.0.1](https://hl7.org/fhir/us/davinci-pas/STU2/qa.html) or [2.2.1](https://hl7.org/fhir/us/davinci-pas/2.2.1/qa.html) versions of the PAS IG QA report. I see "A definition for CodeSystem 'https://www.nubc.org/CodeSystem/RevenueCodes' could not be found, so the code cannot be validated".  Are you actually seeing the error listed here?

karlnaden (changes requested):
This is looking better, and I want to be highly rigorous on the errors that we filter, so a few additional questions. Are these errors showing up on the client vs server suite run using the presets when not suppressed? If not, can you provide an example where they would appear.

lib/davinci_pas_test_kit/cross_suite/validator_suppressions.rb
abhinandan2012: Done. Each extension filter is now anchored to the exact context the validator reports, so the extension is only suppressed at the location the IG forces it - a misplaced use now surfaces. Covered by tests.

lib/davinci_pas_test_kit/cross_suite/validator_suppressions.rb
abhinandan2012: No - it doesn't show up in the per-resource validation the suites rely on (only in bundle-level slice evaluation, where entry issues are already dropped by location), and it's INFO-only. So I removed it from the v2.2.1 list.

lib/davinci_pas_test_kit/cross_suite/validator_suppressions.rb
abhinandan2012: Right - the membership error only appears in my local run (degraded terminology); the authoritative IG QA shows just the CodeSystem-not-found, as you noted. So I dropped both membership suppressions (NUBC and X12278) and kept only CodeSystem-not-found, covered for both revenue-code and CodeSystem/RevenueCodes.

karlnaden: Looks like for NUBC, it is another code system that we can't check the codes for because it is proprietary. Looks like the IG examples don't have any NUBC Revenue Codes because the corresponding elements that hold these codes are must support and not required, which is why we don't see any in the IG QA suppressions.

However, we need to see that element populated for our must support tests, so we'll have to suppress the code validation error - things like the following.

`Claim/ReferralAuthorizationExample: Claim.item[0].revenue: None of the codings provided are in the value set 'AHA NUBC Revenue Value Set' ([http://hl7.org/fhir/us/davinci-pas/ValueSet/AHANUBCRevenueCodes|2.2.1](http://hl7.org/fhir/us/davinci-pas/ValueSet/AHANUBCRevenueCodes%7C2.2.1)), and a coding from this value set is required) (codes = https://www.nubc.org/revenue-code#0120)`

Please add that back.

lib/davinci_pas_test_kit/cross_suite/validator_suppressions.rb
karlnaden: Looks better. One missed based on the preset run: `ClaimResponse.addItem[0]: The extension http://hl7.org/fhir/us/davinci-pas/StructureDefinition/extension-productOrServiceCodeEnd v2.2.1 is not allowed to be used at this point (this element is [BackboneElement, ClaimResponse.addItem]; allowed for this version = e:Claim.item, e:ClaimResponse)` I confirmed that ClaimResponse.addItem is a place that the IG has this extensions, so this error should be added to the supressions.

lib/davinci_pas_test_kit/cross_suite/validator_suppressions.rb
abhinandan2012: Added back - the 'AHA NUBC Revenue Value Set' membership suppression is in the v2.2.1 list.

lib/davinci_pas_test_kit/cross_suite/validator_suppressions.rb
abhinandan2012: Added - productOrServiceCodeEnd is now anchored at both ServiceRequest and ClaimResponse.addItem.

lib/davinci_pas_test_kit/cross_suite/validator_suppressions.rb
```diff
@@ -1,5 +1,61 @@
...
+  #
+  # The v2.2.1 suites match only V221_SUPPRESSED_MESSAGES; the v2.0.1 suites match the union of
+  # both lists (V201_SUPPRESSED_MESSAGES).
+  #
+  # rubocop:disable Layout/LineLength
+  V221_SUPPRESSIONS = [
+    # X12 value set and code system definitions are proprietary and not available to the validator,
+    # so resolution failures for them are not actionable by users.
+    %r{ValueSet 'https://valueset\.x12\.org/[^']+' not found},
+    %r{A definition for CodeSystem 'https://codesystem\.x12\.org/[^']+' could not be found, so the code cannot be validated},
```
karlnaden: I'm seeing warnings like this that these lines aren't catching, e.g., in test "2.1.1.03 Server $submit Response Bundle is conformant": `A definition for CodeSystem 'https://valueset.x12.org/x217/005010/response/2000F/NX1/1/01/00/1345' could not be found, so the code cannot be validated`

karlnaden (changes requested):
one more x12 warning I think should be suppressed

lib/davinci_pas_test_kit/cross_suite/validator_suppressions.rb
abhinandan2012: Thanks for pointing this out. I updated both X12 patterns to support `valueset.x12.org` and `codesystem.x12.org`, and added a test for it.

lib/davinci_pas_test_kit/cross_suite/validator_suppressions.rb
```diff
@@ -1,5 +1,63 @@
...
+    %r{ValueSet 'https://(?:valueset|codesystem)\.x12\.org/[^']+' not found},
+    %r{A definition for CodeSystem 'https://(?:codesystem|valueset)\.x12\.org/[^']+' could not be found, so the code cannot be validated},
+    # The NUBC revenue code system is proprietary as well and cannot be resolved. The test kit's
+    # example data references it under both of the URLs the IG has used for it.
+    %r{A definition for CodeSystem 'https://www\.nubc\.org/(revenue-code|CodeSystem/RevenueCodes)' could not be found, so the code cannot be validated},
+    # Because that code system is unavailable, membership in the PAS-defined value set built on it
+    # cannot be checked either. Claim.item.revenue is must-support (not required), so the IG examples
+    # leave it empty and the IG QA never hits this, but the must support tests populate it, so the
+    # code cannot be checked against the value set and the error is unactionable.
+    "None of the codings provided are in the value set 'AHA NUBC Revenue Value Set' (http://hl7.org/fhir/us/davinci-pas/ValueSet/AHANUBCRevenueCodes",
```
karlnaden, line 27: Two additional PAS-defined value sets with all X12 in them:
- http://hl7.org/fhir/us/davinci-pas/ValueSet/pas-pwk01-attachment-report-type-code
- http://hl7.org/fhir/us/davinci-pas/ValueSet/PASCommunicationRequestMedium

I'm seeing both with warnings in the successful workflow. Let's filter them as well.

karlnaden (changes requested):
found another two to filter

lib/davinci_pas_test_kit/cross_suite/validator_suppressions.rb
abhinandan2012: Added both.

karlnaden (approved):

PR #84, opened 2026-07-16
ID-165: Error handling and cleanup for must support response selection

abhinandan2012:
## Summary

Follow-up cleanup for the criteria-based response selection introduced in #79, addressing code review feedback. The selection semantics are unchanged; these changes harden error handling so that tester input or environment problems produce a logged default response instead of an HTTP 500 to the client under test, and clean up a few internals:

- An invalid or inverted `requestRange` (for example `"1-2,a"` or `"3-2"`) no longer raises. It is logged and treated as unmet, so the entry is skipped and later entries are still considered.
- FHIRPath service failures (non-2xx responses, unreachable service, unparseable response bodies) during criteria evaluation or token replacement are rescued in the endpoint. Inferno logs a warning and generates a default response.
- The operation is now parsed with the query string stripped, so requests to `.../Claim/$submit?param=value` are tagged, counted, and matched correctly.
- When a response input is present but unusable (invalid JSON, or an entry that is neither a Bundle nor a wrapper with a `bundle` key), the reason and offending entry number are logged instead of silently falling back to default responses.
- The request number is computed once per request instead of once per candidate with a `requestRange` criterion, removing redundant repository queries.
- Removed a dead `rescue` in token replacement and skipped the FHIR round trip when the selected bundle contains no tokens.
- The `ms_submit_responses` / `ms_inquire_responses` input descriptions are now generated from a single shared helper in `client_input_descriptions.rb` so the documentation cannot drift between the two inputs.
- FHIRPath service error messages no longer interpolate the full serialized resource.

## How to test

Run the suites against each other following `docs/Running-Suites-Against-Each-Other.md`:
client suite session (Da Vinci PAS Client Suite v2.2.1, Other Authentication, preset
"Run Against the PAS Server Suite") in one tab, server suite session (preset
"Run Against the PAS Client Suite") in another.

### Regression: existing behavior unchanged

1. In the client session, run group `10 Must Support Elements`
   (`pas_client_v221_must_support_no_auth`) and leave test `10.1.01`
   (`pas_client_v221_gather_must_support`) waiting.
2. In the server session, run groups `3.1 $submit Element Support` and
   `3.2 $inquire Element Support`.
3. Back in the client session, click the confirmation link in the `10.1.01` wait dialog
   and complete the attestations.

Expected: the group passes as before, including `10.2.01`
(`pas_client_v221_request_bundle_validation_test`) and `10.2.02`
(`pas_client_v221_response_bundle_validation_test`).

### Error handling scenarios

Run group `10 Must Support Elements` with the `Must Support $submit Response Bundles`
input set to a list whose first entry is a wrapper with an invalid range and whose
second entry is a plain Bundle:

    [
      { "criteria": { "requestRange": "3-2" }, "bundle": { ...bundle A... } },
      { ...bundle B... }
    ]

Send a `$submit` request to the endpoint shown in the `10.1.01` wait dialog. Expected:
HTTP 200 with bundle B returned (the invalid entry is skipped), and the server log
contains "Ignoring unmatchable requestRange criteria".

With the FHIRPath service stopped (`docker stop <fhirpath container>`), provide an
entry with a `fhirpath` criterion and send a request. Expected: HTTP 200 with a
generated default response and a logged warning, rather than an HTTP 500.

Send a `$submit` request with a query string appended to the endpoint URL
(for example `...$submit?param=value`). Expected: the tester-provided bundle is still
returned and the request is counted normally.

Provide input that is not valid JSON, or a wrapper entry missing its `bundle` key.
Expected: default responses, with a log line naming the input and the offending entry.

### Where to verify results

- Test `10.2.01`: the REQUESTS tab lists the received `$submit` requests; open DETAILS
  on a request and expand Response Body to see which bundle was returned.
- The Inferno server log shows one line per request: the selection decisions
  ("Selected tester-provided response bundle N of M...") and the new warnings for
  invalid ranges, unusable inputs, and FHIRPath service failures.
- The run dialog for group `10 Must Support Elements` shows the (now shared) input
  documentation on both response inputs.

### DIFF
--- a/lib/davinci_pas_test_kit/client/endpoints/claim_endpoint.rb
+++ b/lib/davinci_pas_test_kit/client/endpoints/claim_endpoint.rb
@@ -165,14 +165,15 @@ module DaVinciPASTestKit
     # match the incoming request, extracts its response Bundle (unwrapping it when the
     # candidate pairs the Bundle with criteria), and replaces {{fhirpath}} tokens with
     # values from the request. Returns nil, causing Inferno to generate a default
-    # response, if no candidates are provided or none match.
+    # response, if no candidates are provided, none match, or the FHIRPath service
+    # cannot evaluate criteria or tokens.
     def select_must_support_response(req_bundle)
       candidates = UserInputResponse.response_candidates(result, operation)
       return if candidates.blank?
 
       operation_url_suffix = "$#{operation}"
       request_number = count_previous_successful_requests(operation_url_suffix) + 1
-      selected_index = candidates.index { |candidate| include_entity?(candidate, req_bundle, operation_url_suffix) }
+      selected_index = candidates.index { |candidate| include_entity?(candidate, req_bundle, request_number) }
 
       if selected_index.nil?
         Inferno::Application['logger'].info(
@@ -187,16 +188,22 @@ module DaVinciPASTestKit
         "for #{operation_url_suffix} request ##{request_number}."
       )
       replace_tokens(entity_bundle(candidates[selected_index]), req_bundle)
+    rescue FhirpathUtils::FhirpathServiceError, Faraday::Error => e
+      Inferno::Application['logger'].warn(
+        "Unable to select a tester-provided response (#{e.message}). Inferno will generate a default response."
+      )
+      nil
     end
 
-    # Round-trips the selected bundle through the FHIR model to normalize it after
-    # token replacement. Falls back to the replaced string if it no longer parses,
-    # e.g., when a replaced value breaks the JSON structure.
+    # Replaces {{fhirpath}} tokens using values from the incoming request, round-tripping
+    # the result through the FHIR model to normalize it. Falls back to the replaced string
+    # if it no longer parses, e.g., when a replaced value breaks the JSON structure.
     def replace_tokens(bundle_hash, req_bundle)
-      replaced = replace_tokens_in_string(bundle_hash.to_json, req_bundle)
+      bundle_json = bundle_hash.to_json
+      replaced = replace_tokens_in_string(bundle_json, req_bundle)
+      return bundle_json if replaced.equal?(bundle_json)
+
       FHIR.from_contents(replaced)&.to_json || replaced
-    rescue JSON::ParserError
-      replaced
     end
 
     def handle_missing_required_elements(claim_entry, response)
@@ -255,7 +262,8 @@ module DaVinciPASTestKit
     end
 
     def operation
-      request.url&.split('$')&.last
+      operation_with_query = request.url&.split('$')&.last
+      operation_with_query&.split('?')&.first
     end
 
     def start_notification_job(response_bundle_json, decision, generated_claim_response_uuid)
--- a/lib/davinci_pas_test_kit/client/user_input_response.rb
+++ b/lib/davinci_pas_test_kit/client/user_input_response.rb
@@ -24,7 +24,8 @@ module DaVinciPASTestKit
     # JSON array of entries, where each entry is either a bare FHIR Bundle or a
     # wrapper object holding the response Bundle under "bundle" alongside optional
     # selection "criteria".
-    # Returns nil if the input is not present, not parseable, or malformed.
+    # Returns nil, with a log entry when the value is present but unusable, so that
+    # Inferno falls back to generated default responses.
     def self.response_candidates(result, operation)
       input_name = operation == 'submit' ? 'ms_submit_responses' : 'ms_inquire_responses'
       input_value = JSON.parse(result.input_json)&.find { |i| i['name'] == input_name }&.dig('value')
@@ -32,8 +33,20 @@ module DaVinciPASTestKit
 
       candidates = JSON.parse(input_value)
       candidates = [candidates] unless candidates.is_a?(Array)
-      candidates if candidates.all? { |candidate| valid_candidate?(candidate) }
+      invalid_index = candidates.index { |candidate| !valid_candidate?(candidate) }
+      if invalid_index.present?
+        Inferno::Application['logger'].warn(
+          "Ignoring the '#{input_name}' input: entry #{invalid_index + 1} is neither a FHIR Bundle nor a " \
+          'wrapper object with a "bundle" key. Inferno will generate default responses.'
+        )
+        return
+      end
+
+      candidates
     rescue JSON::ParserError
+      Inferno::Application['logger'].warn(
+        "Ignoring the '#{input_name}' input because it is not valid JSON. Inferno will generate default responses."
+      )
       nil
     end
 
--- a/lib/davinci_pas_test_kit/cross_suite/fhirpath_utils.rb
+++ b/lib/davinci_pas_test_kit/cross_suite/fhirpath_utils.rb
@@ -14,8 +14,7 @@ module DaVinciPASTestKit
       return fhirpath_result if fhirpath_result.status.to_s.start_with?('2')
 
       raise FhirpathServiceError,
-            "FHIRPath service returned #{fhirpath_result.status} for query '#{query}' " \
-            "on resource #{body}: #{fhirpath_result.body}"
+            "FHIRPath service returned #{fhirpath_result.status} for query '#{query}'"
     end
 
     def interpret_fhirpath_result_as_boolean(fhirpath_result)
@@ -50,6 +49,8 @@ module DaVinciPASTestKit
         .map { |element| element.is_a?(Array) || element.is_a?(Hash) ? nil : element }
         .compact
         .join(',')
+    rescue JSON::ParserError
+      raise FhirpathServiceError, "FHIRPath service returned an unparseable body for query '#{expression}'"
     end
   end
 end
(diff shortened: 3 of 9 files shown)


abhinandan2012:
Taking this out of draft. It was drafted by mistake during the main vs support-v2.2.0 branch confusion; it was never merged.

Rebased onto current main (now includes ID-88 #83 and ID-151 #94). Same single commit, no code changes. It touches none of the files changed by #99, and specs and rubocop are green on both main and a local merge with the #99 branch.

lib/davinci_pas_test_kit/client/user_input_response.rb
```diff
@@ -24,16 +24,29 @@ def self.user_inputted_response(configurable, operation, result)
...
+          'wrapper object with a "bundle" key. Inferno will generate default responses.'
+        )
+        return
+      end
+
+      candidates
     rescue JSON::ParserError
+      Inferno::Application['logger'].warn(
+        "Ignoring the '#{input_name}' input because it is not valid JSON. Inferno will generate default responses."
+      )
```
karlnaden: logging these to the console is not particularly helpful. I bet we have enough information to log a warning to the test itself where the user will see it. Can you see if that is possible and works?

lib/davinci_pas_test_kit/cross_suite/response_selection_utils.rb
```diff
@@ -82,7 +80,8 @@ def ranges_cover_value?(value, ranges_string)
         end
       end
     rescue ArgumentError => e
-      raise Inferno::Exceptions::TestSuiteImplementationException.new('pas response range criteria', e.message)
+      Inferno::Application['logger'].warn("Ignoring unmatchable requestRange criteria: #{e.message}")
```
karlnaden: look at adding as a warning to the current test, not a console log entry.

karlnaden (changes requested):
I think we want to report errors in the test if possible, not to the console log which the tester won't have access to. Since this is a tester issue in a specific session, it won't be helpful to Inferno maintainers with access to these logs.

lib/davinci_pas_test_kit/client/endpoints/claim_endpoint.rb
```diff
@@ -255,7 +262,8 @@ def make_processing_error_response
     end
 
     def operation
-      request.url&.split('$')&.last
+      operation_with_query = request.url&.split('$')&.last
+      operation_with_query&.split('?')&.first
```
karlnaden: you can use `request.path` to get the  url without the url parameters, I thnk, rather than doing it yourself.

lib/davinci_pas_test_kit/client/endpoints/claim_endpoint.rb
```diff
@@ -187,16 +188,22 @@ def select_must_support_response(req_bundle)
...
+    # if it no longer parses, e.g., when a replaced value breaks the JSON structure.
     def replace_tokens(bundle_hash, req_bundle)
-      replaced = replace_tokens_in_string(bundle_hash.to_json, req_bundle)
+      bundle_json = bundle_hash.to_json
+      replaced = replace_tokens_in_string(bundle_json, req_bundle)
+      return bundle_json if replaced.equal?(bundle_json)
+
       FHIR.from_contents(replaced)&.to_json || replaced
-    rescue JSON::ParserError
-      replaced
```
karlnaden: Why remove the rescue here? FHIR.from_contents could throw an JSON::Parser error, couldn't it? You say "Removed a dead rescue in token replacement and skipped the FHIR round trip when the selected bundle contains no tokens.", but I don't think it is dead.

lib/davinci_pas_test_kit/cross_suite/fhirpath_utils.rb
```diff
@@ -14,8 +14,7 @@ def execute_fhirpath(body, query)
       return fhirpath_result if fhirpath_result.status.to_s.start_with?('2')
 
       raise FhirpathServiceError,
-            "FHIRPath service returned #{fhirpath_result.status} for query '#{query}' " \
-            "on resource #{body}: #{fhirpath_result.body}"
```
karlnaden: The body and the response could be useful for debugging. Maybe those should be logged to the inferno console? Looks like you catch this later and log it to the console. So seems like keeping the body and results here would be a good idea so that it makes it into the Inferno log?

karlnaden (changes requested):

lib/davinci_pas_test_kit/client/endpoints/claim_endpoint.rb
abhinandan2012: Done, switched to request.path.split('$').last.

lib/davinci_pas_test_kit/client/endpoints/claim_endpoint.rb
abhinandan2012: You're right, FHIR.from_contents does raise JSON::ParserError; my "dead" claim was wrong.

lib/davinci_pas_test_kit/cross_suite/fhirpath_utils.rb
abhinandan2012: Restored the service response body in the message, and connection failures now raise the same error with the query and cause.

lib/davinci_pas_test_kit/client/user_input_response.rb
abhinandan2012: Done. response_candidates now raises InvalidInputError with a tester-facing message, and the endpoint records it as a warning on the waiting test's result via the Messages repository. Confirmed the result row survives resume, so the warning shows on 10.1.01. Identical warnings are only recorded once since the input is re-evaluated on every request.

lib/davinci_pas_test_kit/cross_suite/response_selection_utils.rb
abhinandan2012: Done, invalid ranges now add a warning to the waiting test's result instead of a console log entry.

lib/davinci_pas_test_kit/client/user_input_response.rb
```diff
@@ -24,17 +28,25 @@ def self.user_inputted_response(configurable, operation, result)
...
+        raise InvalidInputError,
+              "Entry #{invalid_index + 1} of the '#{input_name}' input is neither a FHIR Bundle nor a " \
+              'wrapper object with a "bundle" key.'
+      end
+
+      candidates
     rescue JSON::ParserError
-      nil
+      raise InvalidInputError, "The '#{input_name}' input is not valid JSON."
     end
```
karlnaden: Since these errors get displayed to the user, they should use the input title, not the name. There should be a way to get the title from the `result.input_json`. See module UserInputResponse's `self.read_input` method for how to get into those details (uses only name and value, but title should be available, too).

Image

lib/davinci_pas_test_kit/client/user_input_response.rb
```diff
@@ -24,17 +28,25 @@ def self.user_inputted_response(configurable, operation, result)
...
       return unless input_value.present?
 
       candidates = JSON.parse(input_value)
       candidates = [candidates] unless candidates.is_a?(Array)
-      candidates if candidates.all? { |candidate| valid_candidate?(candidate) }
+      invalid_index = candidates.index { |candidate| !valid_candidate?(candidate) }
+      if invalid_index.present?
+        raise InvalidInputError,
+              "Entry #{invalid_index + 1} of the '#{input_name}' input is neither a FHIR Bundle nor a " \
+              'wrapper object with a "bundle" key.'
```
karlnaden:
```suggestion
              'wrapper object with a FHIR Bundle in the "bundle" key.'
```

lib/davinci_pas_test_kit/cross_suite/response_selection_utils.rb
```diff
@@ -55,34 +55,50 @@ def request_meets_inclusion_criteria?(expression, request_fhir_obj)
...
 
           (low..high).cover?(value)
         else
           part.to_i == value
         end
       end
     rescue ArgumentError => e
-      raise Inferno::Exceptions::TestSuiteImplementationException.new('pas response range criteria', e.message)
+      add_result_warning(
+        "Ignoring the requestRange criteria #{ranges_string.inspect} because it is not a valid range " \
```
karlnaden:
```suggestion
        "Invalid requestRange criteria #{ranges_string.inspect}. The corresponding Bundle was not selected" \
```

karlnaden (changes requested):
several minor language tweaks

lib/davinci_pas_test_kit/client/user_input_response.rb
abhinandan2012: Done. Note the title isn't in result.input_json on inferno_core 1.4.0 (inputs_as_json only writes name/value/type), so response_candidates now takes the waiting test and reads the title from its input config, falling back to the name. Messages now say 'Must Support $submit Response Bundles'.

Screenshot 2026-09-04 at 9 02 47 PM

lib/davinci_pas_test_kit/client/user_input_response.rb
abhinandan2012: Applied.

lib/davinci_pas_test_kit/cross_suite/response_selection_utils.rb
abhinandan2012: Applied.

karlnaden (approved):

PR #86, opened 2026-07-17
FQM-394: Implement spec-44 tests

kjmahalingam:
# Summary

Adds server-side PAS v2.2.1 coverage for spec-44, which requires each Claim Inquiry Response to either reference the triggering Claim or include a Data Absent Reason explaining why the Claim cannot be referenced.

This PR:

- Verifies each returned inquiry `ClaimResponse` has either:
  - `ClaimResponse.request.reference`, or
  - a `data-absent-reason` extension on `ClaimResponse.request`
- Reads prior `$inquire` responses from Inferno’s tagged request history
- Places the new test under:
  - `Demonstrate Element Support`
    - `$inquire Element Support`
      - `Additional $inquire Response Requirements`
- Updates generated v2.2.1 server suite output and requirement coverage for spec-44

# Testing Guidance

Running the v2.2.1 server suite and client suite against each other using the presets to do so should result in this new test passing. Additionally, the specs / unit tests should pass.


lib/davinci_pas_test_kit/server/v2.2.1/claim_inquiry/server_claim_inquiry_response_tests.rb
kjmahalingam, line 1: The bulk of this was based on the similar functionality for claim updates, but since this is for `$inquire` rather than `$submit` it made sense to create a new code path for it.

lib/davinci_pas_test_kit/server/generated/v2.2.1/pas_server_must_support_group.rb
```diff
@@ -397,6 +398,8 @@ class PASServerMustSupportGroup < Inferno::TestGroup
           test from: :pas_server_v221_inquire_response_must_support_practitioner_role
           test from: :pas_server_v221_inquire_response_must_support_task
         end
+
+        group from: :pas_server_v221_claim_inquiry_responses
```
kjmahalingam, line 402: This is *technically* not a must-support test, but I felt it made sense to put this here since it's the group where `$inquire` responses (including for specifically `ClaimResponse.request`) are already tested. Additionally, doing this makes it easy to continue to use the presets for the client and server suites against one another, since running these groups simultaneously will still allow this test to pass. If we want to move this to a new group we can figure out how best to do that.

lib/davinci_pas_test_kit/generator/server_must_support_group_generator.rb
```diff
@@ -110,6 +110,16 @@ def request_profiles
         ig_metadata.profiles.select { |profile_metadata| MustSupportTargetProfiles.request_profile?(profile_metadata) }
       end
 
+      # Additional Claim Inquiry Response checks are v2.2.x-specific, so they are only included
+      # in the v2.2+ server element support group.
+      def claim_inquiry_responses_group?
```
Jammjammjamm: This seems like a misleading method name.

lib/davinci_pas_test_kit/server/v2.2.1/claim_inquiry/server_claim_inquiry_response_tests.rb
```diff
@@ -0,0 +1,90 @@
...
+      def parsed_response_resource(request)
+        return if request&.response_body.blank?
+
+        FHIR.from_contents(request.response_body)
+      rescue StandardError
+        nil
+      end
+
+      def claim_inquiry_responses
+        response_bundles.flat_map { |bundle| Array(bundle.entry).map(&:resource) }
```
Jammjammjamm: I don't think this is nil-safe, since you haven't checked that `entry.resource` exists.

lib/davinci_pas_test_kit/server/v2.2.1/claim_inquiry/server_claim_inquiry_response_tests.rb
```diff
@@ -0,0 +1,90 @@
...
+      def claim_reference_or_data_absent_reason?(claim_response)
+        claim_reference?(claim_response) || data_absent_reason?(claim_response)
+      end
+
+      def claim_response_label(claim_response, index)
+        claim_response.id.present? ? "ClaimResponse/#{claim_response.id}" : "ClaimResponse at index #{index + 1}"
+      end
+    end
+
+    # spec-44
```
Jammjammjamm: We don't need a comment here since it's specified below.

spec/davinci_pas_test_kit/server/v2.2.1/pas_server_claim_inquiry_response_tests_spec.rb
```diff
@@ -0,0 +1,89 @@
+# Behavioral specs for the server-side Claim Inquiry Response verification tests (spec-44). Inferno
```
Jammjammjamm: I don't tihnk we need comments for specs.

lib/davinci_pas_test_kit/server/v2.2.1/claim_inquiry/server_claim_inquiry_response_tests.rb
kjmahalingam: This is consistent with existing tests. For example: https://github.com/inferno-framework/davinci-pas-test-kit/blob/960e85e7fcd65e2ccc9e1ac97080ca257a3120d7/lib/davinci_pas_test_kit/server/v2.2.1/claim_updates/server_claim_update_response_tests.rb#L84

I can remove this if we want though, and we can make a point of removing the others.

spec/davinci_pas_test_kit/server/v2.2.1/pas_server_claim_inquiry_response_tests_spec.rb
kjmahalingam: This is also consistent with the existing specs: https://github.com/inferno-framework/davinci-pas-test-kit/blob/main/spec/davinci_pas_test_kit/server/v2.2.1/pas_server_claim_update_response_tests_spec.rb#L1-L4

Similarly can remove if we want though.

lib/davinci_pas_test_kit/generator/server_must_support_group_generator.rb
kjmahalingam: Yeah, this one I also based on existing similar code for claim updates: https://github.com/inferno-framework/davinci-pas-test-kit/blob/main/lib/davinci_pas_test_kit/generator/server_suite_generator.rb#L98-L110

But I agree we can make a clearer name, I'll make an update and feel free to suggest a different one if preferred.

lib/davinci_pas_test_kit/server/v2.2.1/claim_inquiry/server_claim_inquiry_response_tests.rb
Jammjammjamm: We shouldn't be adding redundant comments.

spec/davinci_pas_test_kit/server/v2.2.1/pas_server_claim_inquiry_response_tests_spec.rb
Jammjammjamm: Information about the class being tested belongs in that class, not in its spec.

lib/davinci_pas_test_kit/server/v2.2.1/claim_inquiry/server_claim_inquiry_response_tests.rb
```diff
@@ -0,0 +1,89 @@
...
+      def parsed_response_resource(request)
+        return if request&.response_body.blank?
+
+        FHIR.from_contents(request.response_body)
+      rescue StandardError
+        nil
+      end
+
+      def claim_inquiry_responses
+        response_bundles.flat_map { |bundle| Array(bundle.entry).filter_map { |entry| entry&.resource } }
```
Jammjammjamm, line 32: Is it correct to filter out bundles which don't contain any entries, or do we need to require that every bundle contains entries?

kjmahalingam: It's a good question. I think that in practice the answer is that we don't want something like this to happen (see [this page](https://hl7.org/fhir/us/davinci-pas/2.2.1/en/OperationDefinition-Claim-inquiry.html) to show the correct form of a response) but where spec-44 doesn't specifically care about this, it might be better to let profile validation handle such a situation? I suppose it depends on how literally we want the test to conform to spec-44: Should it deliberately and only test the requirements of spec-44 and nothing else, or should it also consider situations like this which should be caught by other tests?

Jammjammjamm (approved):

arscan (approved):

karlnaden (approved):

PR #87, opened 2026-07-23
FQM-393: Add tests for validating echoed resources in $submit responses (spec-33)

kjmahalingam:
# Summary

Adds PAS v2.2.1 coverage for [spec-33](https://hl7.org/fhir/us/davinci-pas/2.2.1/en/conformancedetails.html#ci-c-spec-33).

**Note**: while working on this PR, I also verified the existing functionality for [spec-34](https://hl7.org/fhir/us/davinci-pas/2.2.1/en/conformancedetails.html#ci-c-spec-34).

This PR:

- Activates and updates existing latent validation of echoed resources in PAS $submit responses.
- Ensures an echoed response resource preserves its request `Bundle.entry.fullUrl`, FHIR logical id, and FHIR identifier values.
- Pairs each persisted server response with the request that produced it before validating.
- Adds spec-33 requirement annotations to the v2.2.1 Approval, Denial, and Pended server response-validation tests, and regenerates coverage CSV.

# Testing Guidance

Running the v2.2.1 server suite and client suite against each other using the presets to do so should result in the existing Approval, Denial, and Pended test groups passing. Additionally, the specs / unit tests should pass.


lib/davinci_pas_test_kit/cross_suite/pas_bundle_validation.rb
```diff
@@ -214,7 +214,7 @@ def validate_pa_request_payload_structure(bundle, request_type)
     # a given combination of content. Resources echoed back from the request are validated
     # to ensure the same fullUrl and resource identifiers as in the
     # request are used.
-    def validate_pa_response_body_structure(pa_response_bundle, _pa_request_bundle)
+    def validate_pa_response_body_structure(pa_response_bundle, pa_request_bundle)
```
kjmahalingam, line 217: There was some existing but latent functionality for testing echoed resources that was commented out. I reactivated some of the existing functionality and rewrote some of it to make sure that it conformed to the current expectations of v2.2.1. It may have worked if I uncommented these lines as well, but I also wanted to rewrite some of it and add more validation error messages as possible.

lib/davinci_pas_test_kit/generator/use_case_group_generator.rb
```diff
@@ -111,14 +111,15 @@ def notification_conformance_requirements_verified
...
-        end
+        {
+          ['submit', 'v2.0.1'] => [
+            'hl7.fhir.us.davinci-pas_2.0.1@64', 'hl7.fhir.us.davinci-pas_2.0.1@100',
+            'hl7.fhir.us.davinci-pas_2.0.1@101', 'hl7.fhir.us.davinci-pas_2.0.1@102',
+            'hl7.fhir.us.davinci-pas_2.0.1@103', 'hl7.fhir.us.davinci-pas_2.0.1@107'
+          ],
+          ['submit', 'v2.2.1'] => ['hl7.fhir.us.davinci-pas_2.2.1@spec-33'],
+          ['inquire', 'v2.0.1'] => ['hl7.fhir.us.davinci-pas_2.0.1@131']
+        }[[operation, ig_version]]
```
kjmahalingam, line 124: I added the meaningful line for the new spec, but the rest of the change was to appease rubocop.

lib/davinci_pas_test_kit/cross_suite/pas_bundle_validation.rb
```diff
@@ -224,22 +224,47 @@ def validate_pa_response_body_structure(pa_response_bundle, _pa_request_bundle)
...
+      pa_response_bundle.entry.each do |response_entry|
+        response_resource = response_entry.resource
+        next if response_resource.blank?
+
+        request_entry = request_bundle.entry.find do |entry|
+          echoed_resource?(entry, response_entry)
+        end
+        next if request_entry.blank?
+
+        # Confirm that echoed responses have matching identifiers with their corresponding request
```
Jammjammjamm: If this code needs a comment to explain what it is doing, get rid of the comment and put the code in a method with a name which says what it is doing.

spec/davinci_pas_test_kit/cross_suite/pas_bundle_validation_spec.rb
```diff
@@ -163,6 +163,121 @@ def resource_type
...
+      it 'fails when an echoed resource has a different id' do
+        result = run(
+          test,
+          server_endpoint:,
+          response_body: response_bundle_with_echoed_patient(
+            patient_full_url:, patient_id: 'different-patient', patient_identifier_system:, patient_identifier:
+          ),
+          request_bundle: request_bundle_with_patient
+        )
+        expect(result.result).to eq('fail')
```
Jammjammjamm, line 252: Tests can fail for lots of reasons, so you should add expectations on the failure message to make sure it is failing for the reason being tested in this spec.

lib/davinci_pas_test_kit/cross_suite/pas_bundle_validation.rb
```diff
@@ -224,22 +224,51 @@ def validate_pa_response_body_structure(pa_response_bundle, _pa_request_bundle)
...
-      #     ent.resource.resourceType == res.resourceType && ent.resource.id == res.id
-      #   end
-      #   next unless request_entry.present?
-
-      #   assert(
-      #     request_entry.fullUrl == entry.fullUrl && request_entry.resource.identifier == res.identifier,
-      #     resource_present_in_pa_request_and_response_msg(res)
-      #   )
-      # end
+      validate_echoed_response_resources(pa_response_bundle, pa_request_bundle)
```
Jammjammjamm, line 227: Is this valid for both versions or do we need a version check?

kjmahalingam: Good catch. When reading 2.0.1 requirements, it looks like the same requirement as spec-33 in 2.2.1 is listed at the end of this paragraph [here](https://hl7.org/fhir/us/davinci-pas/STU2/specification.html#prior-authorization-response).

>When echoing back resources that are the same as were present in the prior authorization request, the system SHALL ensure that the same fullUrl and resource identifiers are used in the response as appeared in the request.

So I think this is valid, but if we want to leave 2.0.1 unchanged then we can add a version check in here if that's preferable.

Jammjammjamm (approved):

arscan (approved):

karlnaden (approved):

PR #88, opened 2026-07-23
FQM-395 Spec-55- Servers supporting subscriptions SHALL expose this as part of their CapabilityStatement (CUD)

annagunst:
# Summary
Adds PAS v2.2.1 coverage for [spec-55](https://hl7.org/fhir/us/davinci-pas/2.2.1/en/specification.html#ci-c-spec-55)
>Servers supporting subscriptions SHALL expose this as part of their CapabilityStatement. This includes:
Support for the create interaction
Support for the update interaction
Support for the delete interaction

This PR:
- Created test for spec-55 (pas_server_subscription_capability_statement_test.rb)
- Set up test for spec-55 to be called by pas_server_subscription_setup.rb
- Created spec test/unit test for spec-55 (pas_server_subscription_capability_statement_test_spec.rb)
- Updated davinci_pas_server_suite_v221_requirements_coverage.csv
- Tests that the server has Create AND Updates AND Delete capabilities

# Test Guidance
Running the v2.2.1 server suite and client suite against each other should result in the existing Subscription Setup test groups passing. Additionally, the specs / unit tests should pass.


lib/davinci_pas_test_kit/server/v2.2.1/pas_server_subscription_capability_statement_test.rb
```diff
@@ -0,0 +1,38 @@
...
+
+        subscription_resource = server_rest.resource.find { |entry| entry.type == 'Subscription' }
+        assert subscription_resource.present?,
+               'CapabilityStatement is missing a `Subscription` resource entry in its server-mode `rest` section.'
+
+        supported_interactions = subscription_resource.interaction.map(&:code)
+
+        missing_interactions = REQUIRED_INTERACTIONS - supported_interactions
+
+        assert missing_interactions.empty?, "Missing: #{missing_interactions.join(', ')}"
```
kjmahalingam: Could we make this message more descriptive to indicate that we're missing interactions in the subscription?

build_and_push.sh
kjmahalingam, line 1: This file is not necessary for the global repo so we should remove it from this PR.

docker-compose.background.yml
kjmahalingam, line 1: This file is necessary, so we should return it to its original state so there is no diff on this PR.

docker-compose.yml
kjmahalingam, line 1: Similarly, this file is also necessary, so we should return it to its original state so there is no diff on this PR.

kjmahalingam (changes requested):

Jammjammjamm (approved):

kjmahalingam (approved):
Looks good! I think you just need to address the merge conflicts before it's ready to merge.

kjmahalingam (approved):

kjmahalingam (approved):

Jammjammjamm (approved):

karlnaden (approved):

PR #92, opened 2026-07-29
FQM-404: Add test to verify non-standard information not required

tonydao2:
# Summary
Adds server-side PAS v2.2.1 coverage for conf-11 and conf-12 by verifying that a server can process $submit requests without custom extensions.

These requirements cannot be tested directly because the test suite cannot determine whether a PAS service internally depends on non-standard data or data that is not mandatory or must-support. The server can return a successful response when the request contains only standard FHIR and supported implementation-guide extensions.

This PR:
- Verifies at least on successful $submit request does not contain unsupported custom extensions
- Allows FHIR, US Core, PAS, CRD, DTR, and HRex extensions
- Reads prior $submit requests from Inferno’s tagged request history
- Places the new test under:
     - Demonstrate Element Support
     - $submit Element Support
     - Submission of claims to the $submit operation for must support validation
- Adds requirement coverage for conf-11 and conf-12

# Testing Guidance
Running the v2.2.1 server suite and client suite against each other using the provided presets should result in this new test passing. The relevant specs/unit tests should also pass.


lib/davinci_pas_test_kit/server/v2.2.1/pas_server_no_custom_extensions_test.rb
```diff
@@ -0,0 +1,175 @@
...
+      title 'Server processes $submit without custom extensions'
+      description %(
+        This test verifies that the server is capable of responding to a client
+        without the use of any custom extensions. It inspects each successful
+        PAS submission, and if it finds one which doesn't use any extensions
+        not defined by FHIR, US Core, PAS, CRD, DTR, or HRex it passes.
+      )
+
+      verifies_requirements 'hl7.fhir.us.davinci-pas_2.2.1@conf-11',
+                            'hl7.fhir.us.davinci-pas_2.2.1@conf-12'
```
Jammjammjamm: This should only be testing conf-11.

lib/davinci_pas_test_kit/server/v2.2.1/pas_server_no_custom_extensions_test.rb
```diff
@@ -0,0 +1,174 @@
...
+        'http://hl7.org/fhir/5.0/StructureDefinition/extension-Task.input.value',
+        'http://hl7.org/fhir/5.0/StructureDefinition/extension-Task.output.value',
+        'http://hl7.org/fhir/5.0/StructureDefinition/extension-Task.statusReason'
+      ].freeze
+
+      HREX_EXTENSION_URLS = [
+        'http://hl7.org/fhir/us/davinci-hrex/StructureDefinition/extension-CoverageDavinciWellknownLocation'
+      ].freeze
+
+      DTR_EXTENSION_URLS = [
```
Jammjammjamm, line 41: These are not all of the DTR extensions.

lib/davinci_pas_test_kit/server/v2.2.1/pas_server_no_custom_extensions_test.rb
```diff
@@ -0,0 +1,174 @@
...
+        'http://hl7.org/fhir/us/davinci-pas/StructureDefinition/extension-documentInformation',
+        'http://hl7.org/fhir/us/davinci-pas/StructureDefinition/extension-epsdtIndicator',
+        'http://hl7.org/fhir/us/davinci-pas/StructureDefinition/extension-errorElement',
+        'http://hl7.org/fhir/us/davinci-pas/StructureDefinition/extension-errorFollowupAction',
+        'http://hl7.org/fhir/us/davinci-pas/StructureDefinition/extension-errorPath',
+        'http://hl7.org/fhir/us/davinci-pas/StructureDefinition/extension-homeHealthCareInformation',
+        'http://hl7.org/fhir/us/davinci-pas/StructureDefinition/extension-identifierJurisdiction',
+        'http://hl7.org/fhir/us/davinci-pas/StructureDefinition/extension-identifierSubDepartment',
+        'http://hl7.org/fhir/us/davinci-pas/StructureDefinition/extension-jurisdiction',
+        'http://hl7.org/fhir/us/davinci-pas/StructureDefinition/extension-subDepartmen',
```
Jammjammjamm: Where did you find the `http://hl7.org/fhir/us/davinci-pas/StructureDefinition/extension-jurisdiction` and `http://hl7.org/fhir/us/davinci-pas/StructureDefinition/extension-subDepartmen` extensions?

lib/davinci_pas_test_kit/server/v2.2.1/pas_server_no_custom_extensions_test.rb
```diff
@@ -0,0 +1,174 @@
...
+      ].freeze
+
+      DTR_EXTENSION_URLS = [
+        'http://hl7.org/fhir/us/davinci-dtr/StructureDefinition/activeRole',
+        'http://hl7.org/fhir/us/davinci-dtr/StructureDefinition/information-origin',
+        'http://hl7.org/fhir/us/davinci-dtr/StructureDefinition/intendedUse',
+        'http://hl7.org/fhir/us/davinci-dtr/StructureDefinition/qr-context'
+      ].freeze
+
+      PAS_EXTENSION_URLS = [
```
Jammjammjamm, line 54: The extensions referenced in the [PAS Subscription profile](https://hl7.org/fhir/us/davinci-pas/2.2.1/en/StructureDefinition-profile-subscription.html) need to be included.

lib/davinci_pas_test_kit/server/v2.2.1/pas_server_no_custom_extensions_test.rb
tonydao2: I added those while testing against the Client Suite earlier. At the time, it was not identifying them as custom extensions, and I noticed that `extension-identifierJurisdiction` and `extension-identifierSubDepartment` were already included and thought they were similar. I believe the Client Suite had not yet been updated. I tested it again and it now works without these entries, so I’ll remove them now.

Jammjammjamm (commented):
After spending some more time with DTR, I realized that the extensions used by the SDC IG also need to be included: https://hl7.org/fhir/uv/sdc/STU4/en/index.html

Jammjammjamm (commented):
This should be good, but it needs to be rebased onto the current `support-v2.2.0` branch.

Jammjammjamm (approved):

arscan (approved):

karlnaden (approved):

karlnaden:
fine to merge once merge conflicts resolved.

karlnaden:
@Jammjammjamm @arscan Design question on this type of requirement check. We've taken a similar approach to this kind of requirement but implemented in a significantly different way that I think is worth flagging and discussing before merging this PR.  See https://github.com/inferno-framework/davinci-crd-test-kit/pull/133 for what we are trying. Inferno's use of the validator is by default permissive on unknown extensions (`"extensions": ["any"]` in the validationContext) but can be made more restrictive. Right now, that requires a separate fhir_resource_validator definition, but in [the future might be something that can be defined during the validation call](https://github.com/inferno-framework/inferno-core/issues/748). Threading the validator definition through the shared validation code in CRD was a bit painful, but I think I like it better than trying to maintain whitelisted extensions like this PR does.

It is possible that with the packages loaded this approach taken for CRD will not allow all of the extensions you've specified in here, but it looks like [PAS pulls in all of the referenced IGs](https://hl7.org/fhir/us/davinci-pas/2.2.1/qa.html) (for some version at least) and there are ways to [add extension domains that are allowed](https://confluence.hl7.org/spaces/FHIR/pages/35718580/Using+the+FHIR+Validator#UsingtheFHIRValidator-Extensions) in the validator context.

We do plan on adding our approach to this implementation to this test kit for responses in the next week or two, which would essentially go into the shared PAS Bundle validation code and so be usable by the server tests.  I'm wondering
1. If you agree that this is a better approach, or if you think there are gaps that your approach does cover that this doesn't or other reasons that your approach is preferable.
2. If you do agree our approach is better, would you be willing to wait on this PR and switch over to it once we have that in place, or let us make that change on the server side as a part of our work?

Thanks and let me know if you want to have a further discussion about this.

PR #93, opened 2026-07-31
FQM-425: Implemented spec-47

tonydao2:
# Summary

  Adds server-side PAS v2.2.1 coverage for spec-47.

  This PR:
  - Reuses the must-support `$inquire` request payload supplied by the user.
  - Extracts the authorization number or administration reference number from the request Bundle.
  - Sends the request Bundle to the server’s `Claim/$inquire` endpoint.
  - Parses the returned `Parameters` resource and its PAS Response Bundles.
  - Verifies that the response contains a corresponding reference number that differs from the submitted reference number.
  - Updates the generated requirements coverage for spec-47.
  - Places new test under:
     - Demonstrate Element Support
        - `$inquire` Element Support
        - Submission of claims to the `$inquire` operation for must-support validation

# Testing Guidance
Running the v2.2.1 server suite and client suite's must support against each other using the presets to do so should result in this new test passing.


lib/davinci_pas_test_kit/server/generated/v2.2.1/pas_server_must_support_group.rb
```diff
@@ -369,6 +370,22 @@ class PASServerMustSupportGroup < Inferno::TestGroup
               }
             )
           end
+          test from: :pas_server_v221_replaced_pa_inquire_reference_test do
+            id :pas_server_v221_replaced_pa_inquire_reference_test
+            config(
+              inputs: {
+                replaced_pa_inquire_request_body: {
```
Jammjammjamm: Why is this input being reconfigured?

lib/davinci_pas_test_kit/server/generated/v2.2.1/pas_server_must_support_group.rb
```diff
@@ -369,6 +370,22 @@ class PASServerMustSupportGroup < Inferno::TestGroup
...
+                  name: :must_support_pa_inquire_request_payload,
+                  title: 'Additional $inquire Request Payloads',
+                  description: 'Insert an additional request bundle or a list of bundles (e.g. [bundle_1, bundle_2])'
+                }
+              },
+              options: {
```
Jammjammjamm: What are these options for?

lib/davinci_pas_test_kit/server/v2.2.1/server_replaced_pa_inquire_reference_test.rb
```diff
@@ -0,0 +1,116 @@
...
+            type: 'textarea',
+            description: %(
+              Provide a PAS Inquiry Request Bundle containing an authorization
+              number or administration reference number that is no longer current.
+              The server must be configured to return the current authorization
+              response with a different corresponding reference number.
+            )
+
+      run do
+        request_payload = JSON.parse(replaced_pa_inquire_request_body)
```
Jammjammjamm: This will result in an unhandled exception if the JSON is not valid.

lib/davinci_pas_test_kit/server/v2.2.1/server_replaced_pa_inquire_reference_test.rb
```diff
@@ -0,0 +1,116 @@
...
+        request_payload = request_payload.first if request_payload.is_a?(Array)
```
Jammjammjamm: Why does the input accept an array if only one element will ever be used?

lib/davinci_pas_test_kit/server/v2.2.1/server_replaced_pa_inquire_reference_test.rb
```diff
@@ -0,0 +1,116 @@
...
+        assert request_resource.is_a?(FHIR::Bundle),
+               'The inquiry request body was not a FHIR Bundle.'
+
+        submitted_references = extract_references(request_resource)
+
+        assert submitted_references.values.flatten.present?,
+               'The inquiry request did not contain an authorization number or ' \
+               'administration reference number.'
+
+        fhir_operation(
```
Jammjammjamm: This should be more than one test so that we can separate the parts that validate the user input from the parts that validate the server response. See the existing `$inquire` tests.

lib/davinci_pas_test_kit/server/v2.2.1/server_replaced_pa_inquire_reference_test.rb
```diff
@@ -0,0 +1,116 @@
...
+
+        assert different_reference?(submitted_references, returned_references),
+               'Expected the inquiry response to contain a different authorization number or ' \
+               'administration reference number, but it did not.'
+      end
+
+      def extract_references(bundles)
+        references = Hash.new { |hash, url| hash[url] = [] }
+
+        Array(bundles).each do |bundle|
```
Jammjammjamm: Use `each_with_object` to build up a Hash while iterating over something.

lib/davinci_pas_test_kit/server/v2.2.1/server_replaced_pa_inquire_reference_test.rb
```diff
@@ -0,0 +1,116 @@
...
+          next unless bundle.is_a?(FHIR::Bundle)
+
+          Array(bundle.entry).each do |entry|
```
Jammjammjamm: You shouldn't need to use `Array` to wrap all these fields. `fhir_models` should return an empty array for empty fields with a maximum cardinality greater than one.

```
pry(main)> FHIR::Bundle.new.entry
=> []
```

lib/davinci_pas_test_kit/server/generated/v2.2.1/pas_server_must_support_group.rb
tonydao2: I was basing this off the other $inquire tests in the must support group and saw that it was present in them but can remove.

lib/davinci_pas_test_kit/server/v2.2.1/server_replaced_pa_inquire_reference_test.rb
tonydao2: I assumed the input was an array because the shared must-support inquiry payload supports multiple requests bundles in an array but can change to accept only one element.

spec/davinci_pas_test_kit/server/v2.2.1/server_replaced_pa_inquire_reference_test_spec.rb
```diff
@@ -0,0 +1,129 @@
...
+  let(:input_test) do
+    DaVinciPASTestKit::DaVinciPASV221::PASServerReplacedPAInquireReferenceInputTest
+  end
+
+  let(:test) do
+    Class.new(described_class) do
+      fhir_client { url :server_endpoint }
+      input :server_endpoint
+    end
+  end
```
kjmahalingam: I see that this file tests both the new test files added in this PR. I think it might make sense to have a separate spec file for each of the new added tests. So this file can keep all of the tests for the `:test` fixture, and then a new file can be made for testing the `:input_test` fixture.

spec/davinci_pas_test_kit/server/v2.2.1/server_replaced_pa_inquire_reference_test_spec.rb
```diff
@@ -0,0 +1,129 @@
...
+    bundle = request_bundle('')
+
+    result = run(
+      input_test,
+      pa_inquire_request_body: bundle.to_json
+    )
+
+    expect(result.result).to eq('fail')
+    expect(result.result_message).to include('inquiry requests did not contain an authorization number')
+  end
```
kjmahalingam: It looks like there's only one test for the `:input_test` and it tests a failure case. I think we should add at least another passing test, and potentially other failure cases if there are any other relevant ones. (These can be added in the additional spec file this test should be moved to, per my earlier comment.)

lib/davinci_pas_test_kit/server/v2.2.1/server_replaced_pa_inquire_reference_test.rb
```diff
@@ -0,0 +1,108 @@
...
+      include ParametersHelper
+
+      AUTHORIZATION_NUMBER_EXTENSION =
+        'http://hl7.org/fhir/us/davinci-pas/StructureDefinition/extension-authorizationNumber'.freeze
+      ADMINISTRATION_REFERENCE_NUMBER_EXTENSION =
+        'http://hl7.org/fhir/us/davinci-pas/StructureDefinition/extension-administrationReferenceNumber'.freeze
+      REFERENCE_EXTENSION_URLS = [
+        AUTHORIZATION_NUMBER_EXTENSION,
+        ADMINISTRATION_REFERENCE_NUMBER_EXTENSION
+      ].freeze
```
kjmahalingam: Can we move these constants definitions to a more centralized file? Perhaps `lib/davinci_pas_test_kit/cross_suite/pas_constants.rb` or otherwise another new constants/utils file for `$inquire` reference tests? That way we can reference them in this file and the other new test instead of duplicating them.

kjmahalingam (changes requested):

lib/davinci_pas_test_kit/server/generated/v2.2.1/pas_server_must_support_group.rb
```diff
@@ -369,6 +371,30 @@ class PASServerMustSupportGroup < Inferno::TestGroup
               }
             )
           end
+          test from: :pas_server_v221_replaced_pa_inquire_reference_input_test do
```
Jammjammjamm: These tests are only used in one place, so they shouldn't have to configure their inputs. The correct input definitions should just be in the tests. Configuring inputs is only for when a test is used in multiple places.

lib/davinci_pas_test_kit/server/v2.2.1/server_replaced_pa_inquire_reference_input_test.rb
```diff
@@ -0,0 +1,60 @@
...
+
+module DaVinciPASTestKit
+  module DaVinciPASV221
+    class PASServerReplacedPAInquireReferenceInputTest < Inferno::Test
+      include PASConstants
+
+      id :pas_server_v221_replaced_pa_inquire_reference_input_test
+      title 'Provided $inquire Request contains a replaced prior authorization reference'
+      description %(
+        **USER INPUT VALIDATION**: This test validates tester-provided input.
```
Jammjammjamm: Rather than saying this in text, this test should include the `simulation_verification` method to display this in the UI.

lib/davinci_pas_test_kit/server/v2.2.1/server_replaced_pa_inquire_reference_input_test.rb
```diff
@@ -0,0 +1,60 @@
...
+        skip_if pa_inquire_request_body.blank?,
+                'No inquiry request body was provided.'
+
+        assert_valid_json(
+          pa_inquire_request_body,
+          'Provide a valid JSON PAS Inquiry Request Bundle.'
+        )
+
+        parsed_payload = JSON.parse(pa_inquire_request_body)
+        request_payloads = [parsed_payload].flatten.compact.uniq
```
Jammjammjamm:
```suggestion
        request_payloads = Array.wrap(parsed_payload)
```

lib/davinci_pas_test_kit/server/v2.2.1/server_replaced_pa_inquire_reference_test.rb
```diff
@@ -0,0 +1,101 @@
...
+
+      input :pa_inquire_request_body
+
+      run do
+        assert_valid_json(
+          pa_inquire_request_body,
+          'Provide a valid JSON PAS Inquiry Request Bundle.'
+        )
+        parsed_payload = JSON.parse(pa_inquire_request_body)
+        request_payloads = [parsed_payload].flatten.compact.uniq
```
Jammjammjamm:
```suggestion
        request_payloads = Array.wrap(parsed_payload)
```

lib/davinci_pas_test_kit/server/v2.2.1/server_replaced_pa_inquire_reference_test.rb
```diff
@@ -0,0 +1,101 @@
...
+
+          fhir_operation(
+            '/Claim/$inquire',
+            body: request_resource,
+            name: :pa_inquire_request
+          )
+
+          assert_response_status(200)
+
+          assert resource.is_a?(FHIR::Parameters),
```
Jammjammjamm: I think inquire can return a plain Bundle without wrapping it in Parameters.

spec/davinci_pas_test_kit/server/v2.2.1/server_replaced_pa_inquire_reference_input_test_spec.rb
```diff
@@ -0,0 +1,66 @@
...
+                ]
+              }
+            ]
+          }
+        }
+      ]
+    }
+  end
+
+  it 'fails when the request body does not contain a reference number' do
```
Jammjammjamm, line 36: These specs should test all of the failure modes for these tests.

kjmahalingam (approved):

spec/davinci_pas_test_kit/server/v2.2.1/server_replaced_pa_inquire_reference_input_test_spec.rb
```diff
@@ -0,0 +1,94 @@
...
+      must_support_pa_inquire_request_payload: operation_outcome.to_json
+    )
+
+    expect(result.result).to eq('fail')
+    expect(result.result_message).to include(
+      'Each inquiry request body must be a FHIR Bundle.'
+    )
+  end
+
+  it 'skips when no inquiry request body is provided' do
```
Jammjammjamm: This is base framework functionality. We don't need to unit test it.

spec/davinci_pas_test_kit/server/v2.2.1/server_replaced_pa_inquire_reference_input_test_spec.rb
Jammjammjamm: This spec still isn't testing all of the assertions in the test.

Jammjammjamm (approved):

karlnaden (approved):

